PR #623, opened 2019-03-10
GraphQL Compatibility Acceptance Tests

glucaci:
Implementation of GraphQL Compatibility Acceptance Tests


glucaci:
Focusing for the moment on tests generation, afterwords we will decide how this will be integrated (automatic generation, solution integration)

PR #721, opened 2019-05-07
Added support for XML documentation.

willwolfram18:
### Summary
Code-first schemas use XML doc strings as default descriptions.

- Code-first schemas will now use XML doc strings as default descriptions for the following items
    - Types
    - Fields
    - Enum values
    - Query methods and their parameters
- Users can override the XML doc strings by using the `Description()` methods for describing types, fields, or arguments.

Addresses #715


willwolfram18:
Just took a look at some of the checks, specifically the Sonar code smells. It looks like a majority of the issues are lines that exceed 80 characters, and then a preference of `await` vs. `GetAwaiter().GetResult()`.

I'm happy to fix the line lengths if that's an issue and a hard requirement for merging the PR.

Unfortunately `await` isn't an option since we're in synchronous methods, but I can do some research to figure out whether the preference for making `async` methods synchronous is `GetAwaiter().GetResult()` or `Task.Run().Result` if you want that fixed.

willwolfram18:
Alright, in doing some research regarding using async methods in sync methods, I found this [StackOverflow post](https://stackoverflow.com/a/40344759/5884242) that suggests the best solution is to use `Task.Run(task).GetAwaiter().GetResult()`. This should avoid deadlocks, and provide exception propagation similar to `await`. I will make the change to the branch to properly reflect this.

michaelstaib:
OK, I have looked at your PR now.

We should not depend on another library to achieve this. We try to minimize external dependencies and I would not want to be dependent on something other than framework dependencies.

Furthermore, we should move the description logic into the `DefaultNamingConventions` all the naming and description logic is centralized there. This will make this a much smaller change.

The current implementation there is:

```csharp
public virtual string GetMemberDescription(
            MemberInfo member,
            MemberKind kind)
        {
            if (member == null)
            {
                throw new ArgumentNullException(nameof(member));
            }

            return member.GetGraphQLDescription();
        }
```

This should now be stacked since we should first look for the attribute which is done by `member.GetGraphQLDescription();`.

If that is null then we should have a look at the documentation file since if someone sets the attribute it is more explicit.

willwolfram18:
Oof. If only I had seen that `DefaultNamingConventions` sooner! Would have made things much easier. I'll work on bringing the logic into those classes (and now I feel silly not noticing where the `INamingConventions` is used in those constructors).

I'll work on replacing the NJsonSchema with a self-contained implementation. 👍

src/Core/Types/Types/Descriptors/EnumValueDescriptor.cs
```diff
@@ -1,4 +1,6 @@
 using System;
+using System.Linq;
```
willwolfram18: Reminder to myself to remove these unnecessary `using` statements

src/Core/Types/Types.csproj
```diff
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="GreenDonut" Version="2.0.0" />
+    <PackageReference Include="NJsonSchema" Version="9.13.37" PrivateAssets="True" />
```
willwolfram18: Will replace with a self-contained implementation of the [`XmlDocumentationExtensions`](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs).

michaelstaib:
The naming conventions allow users of HC to overwrite our conventions that we use. So, everything that can be inferred is centralized in the conventions API.

Still, I think this PR will become an awesome addition, and do not feel silly... HC is a big API and the conventions are new with version 9.

src/Core/Types.Tests/Utilities/XmlDocumentationExtensionTests.cs
```diff
@@ -0,0 +1,312 @@
+using System.Drawing;
```
willwolfram18: This file is effectively a copy-paste of [`XmlDocumentationExtensions`](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs) in NJsonSchema, and adapted to use the correct .NET Standard APIs as opposed to dynamic invocation through reflection.

src/Core/Types/Utilities/XmlDocumentationExtensions.cs
```diff
@@ -0,0 +1,514 @@
+using System;
```
willwolfram18: This file is effectively a copy-paste of [`XmlDocumentationExtensions`](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs) in NJsonSchema, and adapted to use the correct .NET Standard APIs as opposed to dynamic invocation through reflection.

src/Core/Types/Utilities/XmlDocumentationExtensions.cs
```diff
@@ -0,0 +1,514 @@
...
+            }
+
+            var value = new StringBuilder();
+            foreach (var node in element.Nodes())
+            {
+                var currentElement = node as XElement;
+                if (currentElement == null)
+                {
+                    value.Append(node);
+                    continue;
```
willwolfram18: I updated this function to use `continue` statements because it had a rather deep nesting structure to accomplish this work. If you're opposed to using `continue` statements and structures, I can revert that change. I found this to be more readable.

Original source: https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs#L138-L181

michaelstaib:
The changes look good so far ... I will have a deeper look at it today.

michaelstaib:
Hi @willwolfram18,

I have started rewriting the xml documentation code. I want it do be cleaner and more efficient. I will separate it in the file logic that reads data, the xml data stuff and the unescape functionality.

I have checked already some in, it will not build anymore but you can already see how I will change it. It will become much more readable and efficient.

michaelstaib:
I think we will get this PR in version 9 next week. I will get you a heads up as soon as I am through with cleaning the xml logic up.

src/Core/Types/Utilities/XmlDocumentationExtensions.cs
michaelstaib: I think this Is better than the original!

src/Core/Types/Utilities/XmlDocumentationExtensions.cs
michaelstaib: I have rewritten it to be more maintainable. I am still working on it.

willwolfram18:
That's exciting to hear this might make the 9.0 release! Let me know if
there's more I can do to help!

Changes look good and make sense! As I'm sure you can tell, I primarily did
a lift-and-shift. I definitely wanted to clean it up more but didn't have a
ton of time. Figured better to get it working first and then refactor from
there 😊

michaelstaib:
Yeah I saw.

While I were on my bus home I thought some more about the xml document implementation and I think I will refactor this class into a service (dependency injection wise). 

At the moment this is static with a static cash. this means that we build one or more schemas and then the cash stays filled. OK, there is an internal clear method on it but with an actual instance class we could just get rid of it.

Also this makes testing with it easier since we can spin up a new instance for each test. Furthermore, by having this as a service we allow people to reimplement it or override parts of it.

willwolfram18:
Making a service sounds like a good call! I will try working on that refactor tomorrow unless you beat me to it 😸

michaelstaib:
> Making a service sounds like a good call! I will try working on that refactor tomorrow unless you beat me to it 😸

Have a go at it :) 

I will try to figure out how we integrate this one so that it plays nicely with the schema builder.

michaelstaib:
I think I am almost through. Just have to fix the tests then we can merge.

michaelstaib:
So, this should done no... if all tests go green.

michaelstaib:
#753

PR #858, opened 2019-06-20
[WIP] File System persisted queries

willwolfram18:
This is a first-pass/WIP for #266 and is by no means ready for merge. I just wanted to take a crack at this and see if what's been bubbling in my head is lining up with what we might be looking for.

Since the UTF-8 parser hasn't yet been implemented, nor has the `IQuery` interface, I just stubbed one out to convey intent. Ultimately I think the read/write interfaces are just responsible for that: read/write the query given some identifier.

If this looks/feels good then I'm happy to start trying to build out the other storage mediums as far as I can until we get UTF-8 parsing done (or at least started that I can work with).


michaelstaib:
> Since the UTF-8 parser hasn't yet been implemented, nor has the IQuery interface, I just stubbed one out to convey intent. Ultimately I think the read/write interfaces are just responsible for that: read/write the query given some identifier.

That is good, I have already started work on this on the client branch. I will merge that soon.

michaelstaib:
> If this looks/feels good then I'm happy to start trying to build out the other storage mediums as far as I can until we get UTF-8 parsing done (or at least started that I can work with).

I think I can merge something tomorrow for you-

src/Core/PersistedQueries.Abstractions/IWriteStoredQueries.cs
```diff
@@ -0,0 +1,19 @@
+using HotChocolate.Execution;
+using System.Threading.Tasks;
+
+namespace HotChocolate.PersistedQueries
+{
+    /// <summary>
+    /// A tool for storing queries to some persistence medium.
+    /// </summary>
+    public interface IStoreQueries
```
michaelstaib: Should we rename than to IWriteStoredQueries .... just to have some symmetry between the names?

src/Core/PersistedQueries.Abstractions/IWriteStoredQueries.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        /// <summary>
+        /// Stores a given query using the given identifier.
+        /// </summary>
+        /// <param name="queryIdentifier">The query identifier.</param>
+        /// <param name="query">The query to store.</param>
+        /// <returns>An asynchronous operation.</returns>
+        Task StoreQueryAsync(string queryIdentifier, IQuery query);
```
michaelstaib: Same goes for this regarding symmetry of names.

willwolfram18: I originally went with `IWriteStoredQueries` and a method of `WriteStoredQueryAsync()` but something felt funny about using two words back-to-back that mean "persist". That said, I'm in no way opposed to going back to `IWriteStoredQueries`. I think there is still some symmetry between the interfaces in that one states "I _store_ queries" and then the other is "I read _stored_ queries" but it is a much looser symmetry than `IRead` and `IWrite`.

michaelstaib:
@willwolfram18 sorry, was completely out for the last two days. I hope I can merge the interfaces over the weekend.

michaelstaib:
@willwolfram18  you should now have access to the new types.

michaelstaib:
Awesome to see progress on this one 👍

willwolfram18:
Yeah sorry it's been a little slow going. Had some life events that took a
little bit of a priority lately.

I'm wondering if you have any feedback on the IQueryFileMap abstraction
currently in use in the file system storage. On the one hand, I think this
makes sense because we don't want the store implementation to be overly
prescriptive in how it identifies the file. But on the other hand, it makes
me wonder if this is too much abstraction, and whether or not it should be
used in other storages (like the Azure Blob storage I've started).

michaelstaib:
I will look at it today or tomorrow....

michaelstaib:
I have implemented a middleware for reading from the storage and one for writing the the storage. Also I have added some presets so that enabling the active query persistence flow is simple.

I will finish this one up tomorrow.

Open TODOs
- [ ] Query middleware order for write 
- [ ] Redis implementation
- [ ] Tests for the storage implementations
- [ ] How do we add queries to the persisted flow

michaelstaib:
OK, this one is done now for V10. We can add more providers with the next version. I will start on the documentation sometime tonight.

PR #1148, opened 2019-10-18
Adds conventions for descriptors

PascalSenn:
- Adds conventions for descirptors
- Add Filter naming conventions #1087


src/Core/Types/Types/Descriptors/Conventions/ConventionRecord.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace HotChocolate.Types.Descriptors
```
PascalSenn: @michaelstaib  This class is only for registering conventions in the service collections. As we need to make a link between an interface and a implementation of said interface we probably need some wrapper

PascalSenn:
@michaelstaib  I think this is ready 👍

src/Core/Types/Types/Descriptors/Conventions/IDescriptorContext.cs
```diff
@@ -10,8 +10,7 @@ public interface IDescriptorContext
 
         ITypeInspector Inspector { get; }
 
-        T GetConvention<T>() where T : IConvention;
-
-        bool TryGetConvention<T>(out T convention) where T : IConvention;
+        T GetConventionOrDefault<T>(T defaultConvention)
```
PascalSenn: Sweet!

PR #1195, opened 2019-11-06
Remove Camel Case Naming Convention

PascalSenn:
Removes camel case naming convetion because its essentially the same as the snake case one


PascalSenn:
I guess the ServiceProvider snaps are also obsolete

PR #1213, opened 2019-11-19
Add conventions for sorting

PascalSenn:
- FIx type in Filter & Sort Middleware
- Extend Middleware Activator by custom paramter
- Add Sorting Conventions  
    - Naming of Asc and Desc
    - Naming of Argument

First half of #1087


src/Core/Utilities/MiddlewareActivator.cs
```diff
@@ -145,5 +155,19 @@ private static ConstructorInfo CreateConstructor(TypeInfo middleware)
                 }
             }
         }
+
+        private static Dictionary<Type, Expression> AddCustomParameter(
+            Dictionary<Type, Expression> customBase,
+            params object[] customParameters)
+        {
+            foreach (var param in customParameters)
```
PascalSenn: @michaelstaib  should i change this to a basic for loop?

src/Core/Types.Sorting/SortOperationKindType.cs
```diff
@@ -1,7 +1,39 @@
...
         : EnumType<SortOperationKind>
     {
+
+        protected override EnumTypeDefinition CreateDefinition(IInitializationContext context)
+        {
+            var definition = base.CreateDefinition(context);
+            var convention = context.DescriptorContext
+                .GetConventionOrDefault(SortingNamingConventionBase.Default);
+
+            foreach (var value in definition.Values)
```
PascalSenn: @michaelstaib  should i change this to a normal for loop?

PascalSenn:
@michaelstaib  With the current implementation of MiddlewareActivator we can only inject custom paramater of a type not an abstraction. If we want to inject abtractions like interfaces we cannot longer work with `params object[]` but rather have to use `KeyValuePairs<Type, object>` or even a `IReadOnlyDictionary<Type, object>`

PR #1252, opened 2019-12-03
fix(Sort): fix multiple sort

### LINKED ISSUE #1251: Sorting doesn't work when using the QueryableSortMiddleware and sorting on two properties
**Describe the bug**
This appears to be an issue with 11.0 preview 58 (possibly earlier) as this error does not happen on 10.3 preview 5.  When using the QueryableSortingMiddleware by using the UseSorting() method and attempting to sort by more than one property in the query, instead of creating a query for sorting by multiple properties, an exception instead is thrown.

**Additional context**
The exception thrown is an InvalidOperationException with the following details:
```
The visited type is invalid. Only InputObjectType is allowed.
   at HotChocolate.Types.Sorting.SortVisitorBase.Enter(ObjectFieldNode node, ISyntaxNode parent, IReadOnlyList`1 path, IReadOnlyList`1 ancestors)
   at HotChocolate.Types.Sorting.QueryableSortVisitor.Enter(ObjectFieldNode node, ISyntaxNode parent, IReadOnlyList`1 path, IReadOnlyList`1 ancestors)
   at HotChocolate.Language.VisitorExtensions.<>c__19`1.<CreateVisitor>b__19_0(ISyntaxNodeVisitor visitor, ISyntaxNode node, ISyntaxNode parent, IReadOnlyList`1 path, IReadOnlyList`1 ancestors)
   at HotChocolate.Language.VisitorExtensions.Enter(ISyntaxNodeVisitor visitor, ISyntaxNode node, ISyntaxNode parent, IReadOnlyList`1 path, IReadOnlyList`1 ancestors)
   at HotChocolate.Language.VisitorExtensions.Accept(ISyntaxNode node, ISyntaxNodeVisitor visitor, IVisitationMap visitationMap, Func`2 defaultAction)
   at HotChocolate.Language.VisitorExtensions.Accept(ISyntaxNode node, ISyntaxNodeVisitor visitor)
   at HotChocolate.Types.Sorting.QueryableSortMiddleware`1.InvokeAsync(IMiddlewareContext context)
   at HotChocolate.Execution.ExecutionStrategyBase.ExecuteMiddlewareAsync(ResolverContext resolverContext, IErrorHandler errorHandler)
```

I believe this is caused do to a bug in QuerySortVisitor in how it interacts with the SortVisitorBase class.  For the SortVisitorBase class, on Enter, it checks the top of the Types stack to ensure it is an InputObjectType, and, if so, adds to the top of the Types and Operations stack.  On Leave, it removes the top item from the Types and Operations stack.  For QuerySortVisitor, on Enter, it first calls Enter for SortVisitorBase, and uses the information to add to the SortOperations queue.  On leave, it performs it's own cleanup, but does not call Leave on SortVisitorBase, thus creating a stack leak, which causes an issue on the next Enter.

I believe this can be resolved by replacing the `return VisitorAction.Continue;` with `return base.Leave(node, parent, path, ancestors);`

PascalSenn:
Resolves #1251
props to @thejaymann


PascalSenn:
I think this test is broken

PR #1253, opened 2019-12-03
feat(Sorting): enable pre sorting in resolver

PascalSenn:
Queryables can now be pre sorted in the resolver

### DIFF
--- a/src/Core/Types.Sorting/QueryableSortVisitor.cs
+++ b/src/Core/Types.Sorting/QueryableSortVisitor.cs
@@ -38,9 +38,11 @@ namespace HotChocolate.Types.Sorting
                 return source;
             }
 
-            IOrderedQueryable<TSource> sortedSource
-                = source.AddInitialSortOperation(
-                    Instance.Dequeue(), _parameter);
+            if (!(source is IOrderedQueryable<TSource> sortedSource))
+            {
+                sortedSource = source.AddInitialSortOperation(
+                      Instance.Dequeue(), _parameter);
+            }
 
             while (Instance.Any())
             {
@@ -99,7 +101,7 @@ namespace HotChocolate.Types.Sorting
             {
                 Instance.Enqueue(
                     new SortOperationInvocation(
-                        (SortOperationKind) sortField.Type.Deserialize(node.Value.Value),
+                        (SortOperationKind)sortField.Type.Deserialize(node.Value.Value),
                         sortField.Operation.Property));
             }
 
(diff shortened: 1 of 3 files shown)


PascalSenn:
I'll check

PascalSenn:
@michaelstaib  IOrderedQueryable behaves not as one might expect. 

```csharp
IQueryable<Foo> t = new Foo []{}.AsQueryable();
t is IOrderedQueryable<Foo>// true
typeof(IOrderedQueryable<Foo>).IsAssignableFrom(t.GetType()) // true
```
## The Problem
https://stackoverflow.com/questions/5071426/how-can-i-tell-if-an-iqueryable-is-an-iorderedqueryable/22012961
> The problem with Entity Framework is that EF DbSet<T> inherits from DbQuery<T> which itself implements IOrderedQueryable<T>! So your statement always returns true if myQueryable is an EF query but not sorted at all. 

https://stackoverflow.com/questions/8507746/determine-whether-an-iqueryablet-has-been-ordered-or-not
>This doesn't seem to work on a collection like dbSet.AsQueryable() as this will always return true on the is IOrderedQueryable check. – Juri Jan 23 '12 at 13:50

## A solution 
>Yes you can inspect the IQueryable.Expression tree to see if it calls any of the OrderBy/ThenBy methods. Expression trees can be examined by deriving a class from ExpressionVisitor.

>There is an internal OrderingMethodFinder in System.Web - which you could adapt. Here's what I came up with:
https://stackoverflow.com/questions/8507746/determine-whether-an-iqueryablet-has-been-ordered-or-not

src/Core/Types.Sorting/QueryableSortVisitor.cs
```diff
@@ -129,5 +136,42 @@ IOrderedQueryable<TSource> sortedSource
...
 
         #endregion
+
+        // Adapted from internal System.Web.Util.OrderingMethodFinder
+        // http://referencesource.microsoft.com/#System.Web/Util/OrderingMethodFinder.cs
+        class OrderingMethodFinder : ExpressionVisitor
+        {
+            bool _orderingMethodFound = false;
+
+            public override Expression Visit(Expression node)
```
PascalSenn: I changed the reference implementation slightly. This way it stops once a ordermethod is found

src/Core/Types.Sorting/QueryableSortVisitor.cs
```diff
@@ -129,5 +136,42 @@ IOrderedQueryable<TSource> sortedSource
...
+                }
+                return base.Visit(node);
+            }
+
+            protected override Expression VisitMethodCall(MethodCallExpression node)
+            {
+                var name = node.Method.Name;
+
+                if (node.Method.DeclaringType == typeof(Queryable) && (
+                    name.StartsWith(nameof(Queryable.OrderBy), StringComparison.Ordinal) ||
```
PascalSenn: I also added `nameof` instead of magic strings

PR #1374, opened 2020-01-13
Change in input parsing from 10.1 to 10.3

nigel-sampson:
We ran into a breaking change moving from 10.1. to 10.3 based on the new input parsing code.

Changing the code to 
``` csharp
descriptor.Field(i => i.Value)
     .Type<NonNullType<EnumType<FileFormat>>>();
```

resolves the issue.

I don't think these are the best long term tests for this issue just how I came across it.

Maybe related but we noticed two other changes in input validation.
- In 10.1 a variable with a field that was not on the input object was ignored, in 10.3 it fails.
- In 10.1 a int input field would accept an string json literal in 10.3 it will not.


nigel-sampson:
Updated with a better unit test.

PR #1599, opened 2020-03-24
Fixed leaf type handling for included ByteArray type

mbauerdev:
Using the newly included ByteArray type caused an error when it was used as a leaf type.

Example:
```
    type Query {
        foo: Foo
    }
    type Foo {
        bar: ByteArray
    }
    query getBars {
        foo {
            bar
        }
    }
```

Caused error:
System.NotSupportedException : Leaf type `ByteArray` is not supported.

Summary of the changes:
- Added ByteArray to list of leaf types


mbauerdev:
Thanks ;-)

PR #1741, opened 2020-04-20
Feature: parallel dataloaders execution

Mr-Dm1try:
Hello!

I think that data loaders should run in parallel.

How they executes now (screenshots are from the JaegerUI):
untitled

How they will be executing with this change:
{A9CAB703-12F3-4B2F-8B80-FA521AFB0201} png

Stacked dataloader calls are also continue working consistently:
{C73F156E-D8A2-4E38-9143-B406C9894557} png


rstaib:
We are already on it. With DataLoader Spec 2.0 we will have one BatchScheduler for all DataLoaders in one request.

Mr-Dm1try:
Ok, thanks for your time then

PR #1848, opened 2020-04-30
implement spatial types

John0x:
Made ParseLiteral virtual and refactored Deserialize to use TryDeserialize

This should now enable us to create the spatial InputObjectType types.

### DIFF
--- a/src/HotChocolate/Core/src/Types/Types/InputObjectType.cs
+++ b/src/HotChocolate/Core/src/Types/Types/InputObjectType.cs
@@ -53,7 +53,7 @@ namespace HotChocolate.Types
                 || literal is NullValueNode;
         }
 
-        public object ParseLiteral(IValueNode literal)
+        public virtual object ParseLiteral(IValueNode literal)
         {
             if (literal is null)
             {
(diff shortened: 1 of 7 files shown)


src/HotChocolate/Core/src/Types/Types/InputObjectType.cs
```diff
@@ -133,19 +133,9 @@ public virtual bool TrySerialize(object value, out object serialized)
...
         {
-            if (serialized is null)
+            if (TryDeserialize(serialized, out object deserialized))
             {
-                return null;
-            }
-
-            if (serialized is IReadOnlyDictionary<string, object> dict)
-            {
-                return _deserialize(dict);
```
steveoh, line 143: is this method still used elsewhere or can it also be removed?

John0x: you mean the Deserialize method?

steveoh: not the method, but the `private Func<IReadOnlyDictionary<string, object>, object> _deserialize`

Issue #1976, opened 2020-05-22
NullableHelper is catching more than just Nullable References


jonmill:
**Describe the bug**

The `src\HotChocolate\Core\src\Types\Utilities\NullableHelper.cs` file is looking for the `System.Runtime.CompilerServices.NullableAttribute` to help with Nullable Reference Types, but is catching other attributes that use the `NullableAttribute` name, causing the whole process to crash since it isn't the Nullable attribute that it's looking for.

**To Reproduce**
1. Add an attribute to a field that has the name `NullableAttribute`, such as the LINQ-To-DB project's `NotNull` attribute
2. Add the ObjectType mapping
3. Start process

**Expected behavior**

The `NullableHelper` file only checks for the built-in `NullableAttribute`

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Win10
 - Version [e.g. 22] 19041.264

**Additional context**

jonmill:
I have a fix locally that unblocks this issue if there's interest

PascalSenn:
Hi @jonmill  
Thank you for reporting this!! 
As you already have a fix, do you mind to create a PR for it?

jonmill:
Sure no problem! My team is currently blocked on this, what's the usual cadence for bug fixes moving into production? :)

PascalSenn:
@jonmill  Release cycle depends on @michaelstaib  
Should be out quickly. I backported it to V10 so you get it also in these prereleases. 
So currently it is under review

jonmill:
Great, thank you!!

jonmill:
Thanks for the help and getting the release out so quickly @PascalSenn and @michaelstaib, it's much appreciated!

PR #1982, opened 2020-05-24
Fixes ambiguous Nullable Attribute  V10

PascalSenn:
Fixes nullability checking to only look for the .NET Framework Nullable attribute

Addresses #1976


PascalSenn:
Backported teh cahgnes requested in #1979

PR #2027, opened 2020-06-05
Add federation directives

dillanmann:

Added the directives for:
- Key
- Provides
- Requires
- External

Also added dummy FieldSet type, but this is still to be fully implemented

Addresses #1631


dillanmann:
Generated initial version of these directives, have marked as draft as the FieldSet scalar is not properly implemented.

Issue #2029, opened 2020-06-06
HotChocolate ASP.NET Core interfieres with Application Insights Request Telemetry


jesulink2514:
When you use HotChocolate.AspNetCore 10.4.3 package it removes the ability to gather Request Telemetry when you use Microsoft Application Insights.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new ASP.NET Core application using API template.
2. Install HotChocolate.AspNetCore  10.4.3 package.
3. Include HotChocolate services using `services.AddGraphQL(SchemaBuilder.New());`
4. Install Microsoft.ApplicationInsights.AspNetCore 2.14.0 package.
5. Include Application Insights services using `services.AddApplicationInsightsTelemetry();`.
6. Start with debugging the application and show the Diagnostics Tools window in Visual Studio 2019.
7. Refresh application opened in the browser and confirm that no Request Telemetry is showed.
8. You can comment `services.AddGraphQL(SchemaBuilder.New());` and repeat step 6 and confirm that now Request Telemetry is showed.

**Expected behaviour**
If you have ApplicationInsights installed, then Request telemetry should be gathered and reported in the Diagnostics Tools window in Visual Studio 2019.

 - OS: Windows 10 Pro 
 - Version build 2004

When Graphql services are included
![](https://i.imgur.com/s33J1fW.png)

When Graphql services are commented 
![](https://i.imgur.com/aXPGsRb.png)

This happens because DiagnosticsListener's are removed from services when Grapqhl registered the Default Pipeline so I've set up a workaround (I've reviewed source code). The hardest part is related to register `QueryExecutionDiagnostics` because is an internal class.

```
Func<IServiceProvider, ISchema> factory = s => SchemaBuilder.New()
                .AddServices(s)
                .AddType<AuthorType>()
                .AddType<BookType>()
                .AddQueryType<Query>()
                .Create();

            var opts = new QueryExecutionOptions { };

            services.AddSingleton(sp =>
            {
                var d = sp.GetRequiredService<DiagnosticListener>();
                var o = sp.GetServices<IDiagnosticObserver>();

                BindingFlags flags = BindingFlags.NonPublic | BindingFlags.Instance;
                CultureInfo culture = null;
                var instantiatedType = Activator.CreateInstance(typeof(QueryExecutionDiagnostics), flags, null, new object[] { d, o }, culture);
                return instantiatedType as QueryExecutionDiagnostics;
            });

[...]

PR #2030, opened 2020-06-06
Don't remove DiagnosticsListener or DiagnosticSource unless TracingPr…

jesulink2514:
This change avoid removing DiagnoticsListener and DianosticsSource unless TracingPreference is set to a value different from TracingPreference.Never. This change is made to avoid problems with other libraries that already use and need DiagnosticsListener and DiagnosticsSource, for example, Microsoft Application Insights, specifically Request Telemetry.

Summary of the changes

- Avoid removing DiagnosticsListener and DiagnosticSource when TracingPreference is different from Never.

Addresses #2029


jesulink2514:
Already fixed in RC.

PR #2294, opened 2020-09-08
Adds a blogpost for filtering

PascalSenn:


steveoh:
I read through this and I am looking forward to getting back into spatial. I should have left a review though. Do you think you can get a native speaker to review this?

PascalSenn:
@steveoh a review would be very much appreciated! i think we should first let @michaelstaib add his few bits :D 
Thanks a lot! 
Btw. i am adding the spatial types back to the project tonight. Had to do some changes :)

PR #2608, opened 2020-11-18
Fixed active persisted query

hundreder:
GetExtensions() method changed to return correct _readonlyExtensions field. This was a cause of active persisted queries not working


michaelstaib (approved):

christianfredh:
Is this fix included in v11? 
Having some issues with active persisted queries currently, but not sure where the problem lies.

hundreder:
> Is this fix included in v11?
> Having some issues with active persisted queries currently, but not sure where the problem lies.

Yes, as a part of this PR
https://github.com/ChilliCream/hotchocolate/pull/2609

PR #2646, opened 2020-11-23
Bugfix/querysyntaxserializer

dejandjenic:
serialization of querydocument without name but with variables omits variables.
persisted queries can't use serialized document, exception "message": "The following variables were not declared: productId." is thrown

### DIFF
--- a/src/Core/Language/Visitors/QuerySyntaxSerializer.cs
+++ b/src/Core/Language/Visitors/QuerySyntaxSerializer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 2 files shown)


michaelstaib:
I think this one was fixed on 11 ... let me check ...

michaelstaib:
https://github.com/ChilliCream/hotchocolate/blob/03c7290f5a5d016536b98cf525b3bce68603928f/src/HotChocolate/Language/src/Language.SyntaxTree/Utilities/SyntaxSerializer.QuerySyntax.cs#L5

dejandjenic:
Cool, could you please accept pr for version 10?

pon, 23. nov 2020. 12:37 Michael Staib <notifications@github.com> je
napisao/la:

michaelstaib:
Sure, but it needs tests showing that the fix works. One test showing serialization with name and variables works and one that it works for the short-hand syntax and another one that shows it works when the name is omitted but variables are defined. Once we have tests I merge it and create a new 10.5.5

src/Core/Language/Visitors/QuerySyntaxSerializer.cs
```diff
@@ -102,12 +102,16 @@ public QuerySyntaxSerializer(bool useIndentation)
             OperationDefinitionNode node,
             DocumentWriter writer)
         {
-            if (node.Name != null)
+            if (node.Name != null || node.VariableDefinitions.Any())
```
michaelstaib, line 105: since node.VariableDefinitions is a list use Count > 0.

michaelstaib (approved):

Issue #2662, opened 2020-11-24
Support int and string in [ID] deserialization


benmccallum:
I know I've created an issue about this previously, but I've thought about this a bit more and it could be as simple as the below line

https://github.com/ChilliCream/hotchocolate/blob/7daa9285ec8b802a493ffe7df3808f16b1fe3a6c/src/HotChocolate/Core/src/Types/Types/Relay/GlobalIdInputValueFormatter.cs#L44

passing through the clr type (`resultType`) through to a new method we'd expose on the `IIdSerializer` of `.Deserialize(object? value, IExtendedType type)` (or just `Type type`) and then folks like myself can implement their own `IIdSerializer` that takes advantage of this extra type information like so:

```csharp
if (type is int && int.TryParse(value, out var intValue))
{
    return value;
}

if (type is string && !IdSerializer.IsPossibleBase64String(value))
{
    return value;
}

// etc for long and Guid...
```

Wanna give me approval and I'll sort that out this weekend? Would simplify a lot of things for us :)

PR #2663, opened 2020-11-24
IIdSerializer to get result type hint

benmccallum:
Allows an `IIdSerializer` implementation to get a hint about what the expected result type is of `IdValue.Value` so other implementations can do polymorphic deserialization.

Addresses #2662

### DIFF
--- a/src/HotChocolate/Core/test/Execution.Tests/DependencyInjection/RequestExecutorBuilderExtensions_IdSerializer.Tests.cs
+++ b/src/HotChocolate/Core/test/Execution.Tests/DependencyInjection/RequestExecutorBuilderExtensions_IdSerializer.Tests.cs
@@ -275,7 +275,7 @@ namespace HotChocolate.Execution.DependencyInjection
                 return "mock";
             }
 
-            public IdValue Deserialize(string serializedId)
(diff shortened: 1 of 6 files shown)


src/HotChocolate/Core/test/Execution.Tests/DependencyInjection/RequestExecutorBuilderExtensions_IdSerializer.Tests.cs
```diff
@@ -275,7 +275,7 @@ public string Serialize<T>(NameString schemaName, NameString typeName, T id)
                 return "mock";
             }
 
-            public IdValue Deserialize(string serializedId)
+            public IdValue Deserialize(string serializedId, Type resultType = null)
```
benmccallum, line 278: This demonstrates that it's a breaking change but netstandard2.0 doesn't have default interface implementations so I guess this means a v12? Or a v11.1 with some clear release notes.

benmccallum: Outdated as now with a default implementation this isn't needed :) The system works!

michaelstaib (approved):

PR #2679, opened 2020-11-26
Website: Syntax highlighting for HTTP requests

rstaib:


omon77:
The link for Schema Stitching is not available for V11 or from main docs: https://hotchocolate.io/docs/stitching

Issue #2917, opened 2021-01-21
UseFirstOrDefault and UseSingleOrDefault cause type errors when used with mutations. 


Kcola:
When a mutation returns an IQueryable and uses the UseFirstOrDefault / UseSingleOrDefault attribute it causes an error. (11.0.9-preview.3) 

**To Reproduce**
Create a mutation resolver that returns an IQueryable with the `[UseFirstOrDefault]` or `[UseSingleOrDefault]` attribute.
```csharp
    [UseFirstOrDefault, UseProjection]
    public IQueryable<Document> AddBook([Service] Repository _repository, Book book){
        _repository.Add(book);
        return _repository.Get<Document>().Where(x=>x.Docid == book.Docid);
    } 
```
**Error**
```logs
      1. Unable to resolve type reference `None: Document`. (HotChocolate.Types.ObjectType<Mutation>)
      
         at HotChocolate.Configuration.TypeCompletionContext.GetType[T](ITypeReference typeRef)
         at HotChocolate.Types.FieldBase`2.OnCompleteField(ITypeCompletionContext context, TDefinition definition)
         at HotChocolate.Types.OutputFieldBase`1.OnCompleteField(ITypeCompletionContext context, TDefinition definition)
         at HotChocolate.Types.ObjectField.OnCompleteField(ITypeCompletionContext context, ObjectFieldDefinition definition)
         at HotChocolate.Types.FieldBase`2.CompleteField(ITypeCompletionContext context)
         at HotChocolate.Types.FieldInitHelper.CompleteFields[TTypeDef,TFieldType,TFieldDef](ITypeCompletionContext context, TTypeDef definition, IReadOnlyCollection`1 fields)
         at HotChocolate.Types.ObjectType.OnCompleteType(ITypeCompletionContext context, ObjectTypeDefinition definition)
         at HotChocolate.Types.TypeSystemObjectBase`1.CompleteType(ITypeCompletionContext context)
         at HotChocolate.Configuration.TypeInitializer.<CompleteTypes>g__CompleteType|32_0(RegisteredType registeredType)
         at HotChocolate.Configuration.TypeInitializer.ProcessTypes(TypeDependencyKind kind, Func`2 action)
         at HotChocolate.Configuration.TypeInitializer.CompleteTypes()
         at HotChocolate.Configuration.TypeInitializer.Initialize(Func`1 schemaResolver, IReadOnlySchemaOptions options)
[...]

PR #2921, opened 2021-01-23
Fixed UseFirstOrDefault missing type reference issue.

PascalSenn:
Fix missing type reference of first or default

Adresses #2917


michaelstaib (commented):
Looks good ... but tests are red

michaelstaib (approved):

Issue #2969, opened 2021-02-01
Adding PhoneNumber Scalar type


wonbyte:
Adding PhoneNumber Scalar type to `HotChocolate.Types.Scalars`
- https://github.com/ChilliCream/hotchocolate/pull/2992

Part of #2927

wonbyte:
Merged

PR #2972, opened 2021-02-01
Add PhoneNumber scalar

wonbyte:
- Add PhoneNumber Scalar

Addresses #2969


wonbyte:
initial wip, the file size should reduce once #2950 is merged as the PR is based off of it

src/HotChocolate/Core/src/Types.Scalars/PhoneNumberType.cs
```diff
@@ -0,0 +1,114 @@
...
+
+namespace HotChocolate.Types.Scalars
+{
+    /// <summary>
+    /// A field whose value conforms to the standard E.164 format as specified
+    /// in: https://en.wikipedia.org/wiki/E.164.
+    /// </summary>
+    public class PhoneNumberType : StringType
+    {
+        private static readonly string _validationPattern = @"/^\+[1-9]\d{1,14}$/";
```
wonbyte, line 12: @PascalSenn @michaelstaib How do you prefer to handle `PhoneNumber` validation? This is the regex i got from https://github.com/Urigo/graphql-scalars/blob/master/src/scalars/PhoneNumber.ts#L3 but maybe it would be better to use https://github.com/twcclegg/libphonenumber-csharp ?

wonbyte:
will re-open as well

PR #2986, opened 2021-02-01
Added support for more value types to filtering defaults

jfheins:
Some people use uints or other uncommon types in their data models. In my case, we have a number that can only be positive (usually between 0 and 12) and we made it an uint.

Ideally, I want to tell HC once to treat that uint as IntType.

To add some more value, I also included the rest o the basic types.


src/HotChocolate/Data/src/Data/Filters/Convention/Extensions/FilterConventionDescriptorExtensions.cs
```diff
@@ -82,7 +82,11 @@ public static class FilterConventionDescriptorExtensions
                 .BindRuntimeType<DateTimeOffset?,
                     ComparableOperationFilterInputType<DateTimeOffset?>>()
                 .BindRuntimeType<TimeSpan?, ComparableOperationFilterInputType<TimeSpan?>>();
-
+                .BindRuntimeType<sbyte, ComparableOperationFilterInputType<sbyte>>()
```
PascalSenn, line 85: Can you also add the nullables :) And also, they have a order, first the non nullable then the nullable
```csharp
 .BindRuntimeType<sbyte, ComparableOperationFilterInputType<sbyte>>()
 .BindRuntimeType<ushort, ComparableOperationFilterInputType<ushort>>()
 .BindRuntimeType<uint, ComparableOperationFilterInputType<uint>>()
 .BindRuntimeType<ulong, ComparableOperationFilterInputType<ulong>>()
....
 .BindRuntimeType<TimeSpan?, ComparableOperationFilterInputType<TimeSpan?>>();
 .BindRuntimeType<sbyte?, ComparableOperationFilterInputType<sbyte?>>()
 .BindRuntimeType<ushort?, ComparableOperationFilterInputType<ushort?>>()
 .BindRuntimeType<uint?, ComparableOperationFilterInputType<uint?>>()
 .BindRuntimeType<ulong?, ComparableOperationFilterInputType<ulong?>>()
```

PascalSenn (changes requested):
Hi there

Thanks for the PR!

Small change, could you also add the nullable versions of the types?

src/HotChocolate/Data/src/Data/Filters/Convention/Extensions/FilterConventionDescriptorExtensions.cs
jfheins: Sure :-)
I also refactored the code a tiny bit to make it more obvious that all comparable types have also their nullable counterpart added. (But I can also revert that back if you don't like it)

PascalSenn (approved):
Sorry forgot about this one!

PR #2997, opened 2021-02-05
Fixed AnyType for schema stitching

### LINKED ISSUE #2942: Gateway stitching  - AnyType is transformed in the gateway
AnyType is transformed by the gateway from:
```
            "options": {
              "setValues": true,
              "description": "bla bla"
            }
```            
into:
```
            "options": {
              "setValues": {
                "kind": "BooleanValue",
                "location": null,
                "value": true
              },
              "description": {
                "kind": "StringValue",
                "location": null,
                "value": "bla bla",
                "block": false
              }
```
**Version:**
Gateway 11.0.8
Downstream service, reproduced using 10.5.5 and 11.0.8

**Expected behavior**
Gateway should not transform the response.

Attached repro:
[WebApplication2.zip](https://github.com/ChilliCream/hotchocolate/files/5874504/WebApplication2.zip)

PascalSenn:
In stitching we receive ValueNodes from the remote schema
The AnyType has to deserialize these to runtime values. 
We could also leave the serialization away and just handle these cases in deserialization and never properly deserialize
This would be more efficient, but the dictionary could not be evaluated in middlewares & co on the gateway.
Not sure what the better tradeoff is, maybe we could add a flag to the constructor? with the more efficent way?

@michaelstaib should we merge this one in as is and revisit it in 11.2?

Closes #2942. #2691,


src/HotChocolate/Core/src/Types/Types/Scalars/AnyType.cs
```diff
@@ -218,8 +219,55 @@ public override bool TrySerialize(object? runtimeValue, out object? resultValue)
...
+                        else
+                        {
+                            return false;
+                        }
+
+                    }
+                    runtimeValue = result;
+                    return true;
+                }
+
```
PascalSenn, line 262: line?

src/HotChocolate/Core/src/Types/Types/Scalars/AnyType.cs
```diff
@@ -218,8 +219,55 @@ public override bool TrySerialize(object? runtimeValue, out object? resultValue)
...
+                case IValueNode literal:
+                    runtimeValue = ParseLiteral(literal);
+                    return true;
+
```
PascalSenn, line 266: line?

michaelstaib (dismissed):

michaelstaib (changes requested):
I will have a go at this fix

michaelstaib (approved):

PR #2999, opened 2021-02-05
Added support for filtering on relay id.

### LINKED ISSUE #2971: How do you use the relay node ID for filtering?
Assuming I have the following types where the `Id`s implements relay node.

```c#
class Track
{
    public int Id { get; set; }
}
class Session
{
    public int TrackId { get; set; }
}
```

Is there a way to use the `Track.Id` from 

```graphql
query {
  tracks {
    id
  }
}
```

to filter `Session` by `TrackId` like so:

```
query {
  sessions (where: { trackId: { eq: "VHJhY2sKaTI=" } }) {
    ...
  }
}
```

The query above currently throws an error because it expects an `int` instead of an ID/string that is returned in `query { tracks { id } } `

## Use case

In our app, users will have the ability to filter `sessions` by selecting other related entities (e.g. `track`, `location`, etc). As mentioned, there are a number of related entities and we'd like to use them together so `GetSessionByTrackId`, `GetSessionByLocaionId` will not work. Without implementing relay node, we could directly use the related entities' `id`s to filter the `sessions`.

In other words, we're looking into a similar graphql query that will behave like this c# code

```c#
sessions
.Where(s => s.TrackId == trackId)
.Where(s => s.LocationId == locationId)
// ... more filter for other related entities if the user have selected it
```

Thank you!

PascalSenn:
Adds support for filtering on relay id

Closes #2971


src/HotChocolate/Core/src/Types/Types/Relay/GlobalIdInputValueFormatter.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace HotChocolate.Types.Relay
 {
-    internal class GlobalIdInputValueFormatter : IInputValueFormatter
+    public class GlobalIdInputValueFormatter : IInputValueFormatter
```
PascalSenn, line 13: @michaelstaib  Would this be ok?

src/HotChocolate/Data/src/Data/Filters/Types/RelayIdFilterFieldExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        private static void AddSerializerToInputField<TEntityType, TIdType>(
+            ITypeCompletionContext completionContext,
+            FilterOperationFieldDefinition definition)
+        {
+            ITypeInspector typeInspector = completionContext.TypeInspector;
+            IExtendedType? resultType;
+
+            resultType = typeInspector.GetReturnType(typeof(TIdType[]), true);
```
PascalSenn, line 35: This works but I am not sure if it is a hack
Without this it ListValues are not supported.

src/HotChocolate/Data/src/Data/Filters/Types/IdOperationFilterInputType.cs
```diff
@@ -0,0 +1,75 @@
+using HotChocolate.Configuration;
+using HotChocolate.Internal;
+using HotChocolate.Types;
+using HotChocolate.Types.Descriptors.Definitions;
+
+namespace HotChocolate.Data.Filters
+{
+    public class IdOperationFilterInputType<TEntityType, TIdType>
```
PascalSenn, line 8: To initialise the ID serialiser I needed the entity type and the type of the id.
If you have a better idea let me know.
We could push it down with a type interceptor, but in the end, the user has to define the ID operation types anyway by hand, because we do not yet have a way to infere the ID property from the dotnet type.

shidcordero:
when can this be merge to main branch and release? I need it badly

shidcordero:
any update on this one?

src/HotChocolate/Data/src/Data/Filters/Types/RelayIdFilterFieldExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+            IExtendedType resultType,
+            ExtendedTypeReference extendedTypeReference)
+        {
+            NameString typeName =
+                completionContext.GetType<IType>(extendedTypeReference).TypeName();
+            IIdSerializer serializer =
+                completionContext.Services.GetService<IIdSerializer>() ??
+                new IdSerializer();
+
+            return new GlobalIdInputValueFormatter(
```
PascalSenn, line 57: @michaelstaib  i cannot make the constructor internal as i use it here. 
How far are you and @benmccallum  on the input value formatter? can we then make it public?

benmccallum: I've got no objection to it being public, I could unit test it then for the stuff I added.

My PR is done, just waiting on a review and merge; though my tests add a type interceptor and a lot of stuff that's irrelevant to what I'm testing. If this was public I'd simply all that.

PascalSenn:
@michaelstaib  @benmccallum  what is the stage of  `GlobalIdInputValueFormatter`? can we progress with this PR?

braidenstiller:
Any updates on this?

PascalSenn:
Replaced by #4390

Issue #3029, opened 2021-02-11
Authorization always returns AUTH_NOT_AUTHENTICATED


Anton0123:
**Describe the bug**
If you add aspnetcore jwt authorization with the hotchocolate 11 Authorization attribute it will always return unauthorized even if you provide a valid bearer token.

**To Reproduce**
1. Check out the example repository https://github.com/Anton0123/HotchocolateAuthExample
2. Run the project
3. Copy the generated token printed in the console
4. When making a graphql request for "auth", add the http header `"Authorization": "Bearer <token from step 3>"`

**Expected behavior**
"Secret" should be returned.

**Desktop (please complete the following information):**
 - OS: Microsoft Windows 10 Pro
 - Version	10.0.18363 Build 18363

pasqualirb:
For now, I circumvented the problem by using **`AddHttpRequestInterceptor()`** extension method to explicitly decode the token and set property `User` of `HttpContext`.

    services
  
        .AddGraphQLServer()
  
        .AddAuthorization()
  
        .AddHttpRequestInterceptor((context, executor, builder, cancellationToken) => {
            try {
                // Decode token
                var authHeader = context.Request.Headers.Single(p => p.Key == "Authorization");
                var token = (new JwtSecurityTokenHandler()).ReadJwtToken(authHeader.Value);
                // Get username and roles
                var username = token.Claims.Single(p => p.Type.Equals("username")).Value;
                var roles = token.Claims.Where(p => p.Type.Equals("role")).Select(p => p.Value).ToArray();
                // Set User identity
                context.User = new GenericPrincipal(new GenericIdentity(username), roles);
            } catch(Exception ex) { }
            return ValueTask.CompletedTask;
        });

For coming up with this solution I had to debug `TryGetAuthenticatedPrincipal()` and `IsInAnyRole()` of [`AuthorizeMiddleware.cs`](https://github.com/ChilliCream/hotchocolate/blob/11.1.0-preview.59/src/HotChocolate/AspNetCore/src/AspNetCore.Authorization/AuthorizeMiddleware.cs).

Try/catch is used to prevent triggering an exception when, for example, there is no Authorization header.

Anton0123:
> For now, I circumvented the problem by using **`AddHttpRequestInterceptor()`** extension method to explicitly decode the token and set property `User` of `HttpContext`.
> 
> ```
> services
> 
>     .AddGraphQLServer()
> 
>     .AddAuthorization()
> 
[...]

PR #3030, opened 2021-02-11
fix: nullability issue in projections

PascalSenn:
Addresses #3028


Anton0123:
> Addresses #3029

I think you tagged the wrong issue. This looks more like #3028

Issue #3034, opened 2021-02-11
Add PostalCode scalar


wonbyte:
Add `PostalCode` scalar type to `Hotchocolate.Types.Scalars`

- https://github.com/ChilliCream/hotchocolate/pull/3046

Address #2927

PascalSenn:
Awesome!

PR #3046, opened 2021-02-14
Add PostalCode Scalar

wonbyte:
Add `PostalCode` scalar type to `Hotchocolate.Types.Scalars`

Addresses #3034


src/HotChocolate/Core/src/Types.Scalars/PostalCodeType.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        /// <summary>
+        /// Start with a limited set as suggested here:
+        /// http://www.pixelenvision.com/1708/zip-postal-code-validation-regex-php-code-for-12-countries/
+        /// and here:
+        /// https://stackoverflow.com/questions/578406/what-is-the-ultimate-postal-code-and-zip-regex
+        /// </summary>
+        private static readonly Func<string, bool> _validPostalCode = input =>
+            _validationPatterns.Select(pattern => Regex.Match(
+                input, pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase)).Any(match => match.Success);
```
PascalSenn, line 61: I dont this RegexOptions.Compiled has any effect on Regex.Match. 
https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options

src/HotChocolate/Core/src/Types.Scalars/PostalCodeType.cs
```diff
@@ -0,0 +1,156 @@
...
+        };
+
+        /// <summary>
+        /// Start with a limited set as suggested here:
+        /// http://www.pixelenvision.com/1708/zip-postal-code-validation-regex-php-code-for-12-countries/
+        /// and here:
+        /// https://stackoverflow.com/questions/578406/what-is-the-ultimate-postal-code-and-zip-regex
+        /// </summary>
+        private static readonly Func<string, bool> _validPostalCode = input =>
+            _validationPatterns.Select(pattern => Regex.Match(
```
PascalSenn, line 60: In `hotpaths` like scalars we try to avoid Linq as it allocated objects every time it is used.
We even try to avoid foreach loops because they create a enumerator
the best thing is always a for loop

PascalSenn (commented):
Very Nice! 

I will quickly align a few things and merge it

Issue #3050, opened 2021-02-14
Add UnsignedInt scalar type


wonbyte:
Add `UnsignedInt` scalar type to `Hotchocolate.Types.Scalars`
- #3052 

Addresses #2927

PascalSenn:
Thanks for the contribution!!

PR #3052, opened 2021-02-14
Add UnsignedInt scalar

wonbyte:
Add `UnsignedInt` scalar type to `Hotchocolate.Types.Scalars`

Addresses #3050


PascalSenn:
@wonbyte  this looks already good! 
There was one thing missing. 
uint is not the same primitive as int, so we need to extend the intvaluenode 👍  
Will merge this later today!

PascalSenn (approved):

PR #3142, opened 2021-02-28
Added InMemory connection to straw berry shake

PascalSenn:


src/StrawberryShake/Client/src/Core/Extensions/IDocumentExtensions.cs
```diff
@@ -0,0 +1,10 @@
+using System.Text;
+
+namespace StrawberryShake
+{
+    public static class DocumentExtensions
+    {
+        public static string Print(this IDocument document) =>
+            Encoding.UTF8.GetString(document.Body.ToArray());
```
PascalSenn, line 8: @michaelstaib  is there a better way? net standard 2.0 does not support IReadOnlySpan. Should we use pragmas?

Issue #3175, opened 2021-03-03
Refactor existing Regex scalars


wonbyte:
Refactor existing Regex scalars to extend the RegexType

Addresses #2927

PR #3184, opened 2021-03-03
Refactor existing Regex scalars to use Regex Base Type

wonbyte:
Refactor existing Regex scalars in `Hotchocolate.Types.Scalars` to inherit from the `RegexType`

Addresses #3175


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "5.0.100"
-    
+    "rollForward": "latestMinor"
```
wonbyte, line 4: @michaelstaib should I leave this?

PascalSenn (dismissed):

src/HotChocolate/Core/src/Types.Scalars/PostalCodeType.cs
```diff
@@ -144,5 +144,46 @@ private static bool ValidatePostCode(string postCode)
 
             return false;
         }
+
+        private static class PostalCodePatterns
```
wonbyte, line 148: 👍

Issue #3212, opened 2021-03-07
Add RGB Scalar


wonbyte:
Add `RgbType` to `Hotchocolate.Types.Scalars`

- https://github.com/ChilliCream/hotchocolate/pull/3213

Addresses: #2927

PascalSenn:
and again thanks for the contribution!!

PR #3213, opened 2021-03-07
RGB scalar type

wonbyte:
Add `RgbType` to `Hotchocolate.Types.Scalars`

Addresses #3212


wonbyte:
@PascalSenn the `regex` from https://github.com/Urigo/graphql-scalars is only supporting the `functional syntax` and not any of the other variations https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#rgb_syntax_variations. I think this should be ok for now

PascalSenn (approved):
Nicely done!

Issue #2927, opened 2021-01-24
Introducing HotChocolate.Types.Scalars


PascalSenn:
By using scalar types for leaf fields, you can build expressive and easy to understand schemas.

the Guild implemented a very useful set of scalars in JavaScript. We want to port these scalars to HotChocolate and make them available for you.
We already unsealed all scalar types and created a test base that helps us test new scalars.

We have a many people in the community that would like to contribute to HotChocolate but do not know where to start. This issue is an incredibly good starter issue.

# The Scalars

1. [x] [scalar Time](https://github.com/Urigo/graphql-scalars#time)
1. [x] scalar DateTime
1. [x] [scalar Duration](https://github.com/Urigo/graphql-scalars#duration)
1. [x] [scalar UtcOffset](https://github.com/Urigo/graphql-scalars#utcoffset)
1.  [x] [scalar LocalDate](https://github.com/Urigo/graphql-scalars#localdate)
1.  [x] [scalar LocalTime](https://github.com/Urigo/graphql-scalars#localtime)
1.  [x] [scalar LocalEndTime](https://github.com/Urigo/graphql-scalars#localendtime)
1.  [x] [scalar EmailAddress](https://github.com/Urigo/graphql-scalars#emailaddress)
1.  [x] [scalar NegativeFloat](https://github.com/Urigo/graphql-scalars#negativefloat)
1.  [x] [~~scalar NegativeInt~~](https://github.com/Urigo/graphql-scalars#negativeint) Issues #2939 PR #2940
1.  [x]  [~~scalar NonEmptyString~~](https://github.com/Urigo/graphql-scalars#nonemptystring)
1.  [x] [~~scalar NonNegativeFloat~~](https://github.com/Urigo/graphql-scalars#nonnegativefloat)Issues #3013 PR #3015
1.  [x] [scalar NonNegativeInt](https://github.com/Urigo/graphql-scalars#nonnegativeint)
1.  [x] [scalar NonPositiveFloat](https://github.com/Urigo/graphql-scalars#nonpositivefloat)
1. [x]  [scalar NonPositiveInt](https://github.com/Urigo/graphql-scalars#nonpositiveint)
1.  [x] [scalar PhoneNumber](https://github.com/Urigo/graphql-scalars#phonenumber)
1.  [x] [~~scalar PositiveFloat~~](https://github.com/Urigo/graphql-scalars#positivefloat) Issues #2928 PR #2929
[...]

PR #3268, opened 2021-03-13
Added LocalDate scalar

damikun:
Adding LocalDate scalar + Tests

Part of #2927


PascalSenn:
Hi @damikun  

Thanks for the PR! 
I see you chose the Regex type as the base for the LocalDate scalar. I wonder if it would make more sense to have it resolved into a DateTime rather than a string. 
Just discussed something similar with @wonbyte  over in #3281 . What do you think?

damikun:
Okay I can do that... I used string approach since RFC define only format and our source materials use also string... but I will rewrite to use DateTtime parser and use c# runtime.. Tests can be reused... I will ping you in day-two when i find time :P

damikun:
Just for info... I need to add more tests + make it convertible between Offset and DT by default..

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+using HotChocolate.Language;
+
+namespace HotChocolate.Types.Scalars
+{
+    /// <summary>
+    /// The `LocalDate` scalar type represents a ISO date string, represented as UTF-8
+    /// character sequences YYYY-MM-DD. The scalar follows the specification defined in
+    /// <a href="https://tools.ietf.org/html/rfc3339">RFC3339</a>
+    /// </summary>
+    public class LocalDateType  : ScalarType<DateTime, StringValueNode>
```
PascalSenn, line 14:
```suggestion
    public class LocalDateType : ScalarType<DateTime, StringValueNode>
```

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+        private const string _localFormat = "yyyy-MM-dd";
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LocalDateType"/> class.
+        /// </summary>
+        public LocalDateType()
+            : this(
+                WellKnownScalarTypes.LocalDate,
+                ScalarResources.LocalDateType_Description,
+                BindingBehavior.Implicit)
```
PascalSenn, line 25: We need to register this type by default as explicit
```suggestion
                ScalarResources.LocalDateType_Description)
```

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        public override IValueNode ParseResult(object? resultValue)
+        {
+            return resultValue switch
+            {
+                null => NullValueNode.Default,
+                string s => new StringValueNode(s),
+                DateTimeOffset d => ParseValue(d),
+                DateTime dt => ParseValue(new DateTimeOffset(dt)),
```
PascalSenn, line 48: Just use ParseValue for DateTime too, see blow

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+        public override bool TrySerialize(object? runtimeValue, out object? resultValue)
+        {
+            switch (runtimeValue)
+            {
+                case null:
+                    resultValue = null;
+                    return true;
+                case DateTimeOffset dt:
+                    resultValue = Serialize(dt);
+                    return true;
```
PascalSenn, line 77: Add a case for DateTime too

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+                case string s when TryDeserializeFromString(s, out DateTime? d):
+                    runtimeValue = d;
+                    return true;
+                case DateTimeOffset:
+                    runtimeValue = resultValue;
+                    return true;
+                case DateTime dt:
+                    runtimeValue = new DateTimeOffset(
+                        dt.ToUniversalTime(),
+                        TimeSpan.Zero);
```
PascalSenn, line 100: Just return datetime

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+            switch (resultValue)
+            {
+                case null:
+                    runtimeValue = null;
+                    return true;
+                case string s when TryDeserializeFromString(s, out DateTime? d):
+                    runtimeValue = d;
+                    return true;
+                case DateTimeOffset:
+                    runtimeValue = resultValue;
```
PascalSenn, line 95:
```suggestion
                case DateTimeOffset d:
                    runtimeValue = d.DateTime;
```

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
```diff
@@ -0,0 +1,134 @@
...
+                        dt.ToUniversalTime(),
+                        TimeSpan.Zero);
+                    return true;
+                default:
+                    runtimeValue = null;
+                    return false;
+            }
+        }
+
+        private static string Serialize(DateTimeOffset value)
```
PascalSenn, line 108: The trick to handle DateTime and DateTimeOffset is:
```suggestion
        private static string Serialize(IFormattable value)
```

PascalSenn (changes requested):
Hi @damikun  This looks really good!!!!

src/HotChocolate/Core/src/Types.Scalars/LocalDateType.cs
damikun: This is interesting i never heard abut IFormattable

PascalSenn (approved):

Issue #3228, opened 2021-03-09
Adds Currency Scalar Types 


xGuildart:
Adding Currency scalar type to Hotchocolate.Types.Scalars

#2927

PR #3277, opened 2021-03-14
Pull request for Adding Currency scalar type

aouimar:
Adding Currency Scalar type

The currency scalar type Inherite directly from the stringType

Addresses #3228


PascalSenn:
Hi @aouimar 

This diff looks a lot better :) 
I see you parse the currency scalar very accurate. Which is good in one way, but may not be ideal in another. 
I think we could make this implementation a lot simpler by just using the RegexType as a base type like in e.g. `HexColorCode` 
A regex like this would probably be enough: `^([0-9. ,]*[A-Z]{3}|[A-Z]{3}[0-9., ]*)$` 

 You can see an example here: https://regex101.com/r/OhyHXL/1
 
 This is a lot less accurate than you solution. But if I think the complexity is not worth it in this case. As we anyway loos the kind (USD, CAD, ...) and return a string to the user, the user has to parse it any way. Even if we parse it very accurately, the user has to parse it again a second time. 

What do you think?

aouimar:
Hi Pascal,

In my point of view, if we define it as regextype with simple regex by
allowing all codes, we will diverge on iso-4217,
and the user will pass any currency, bitcoins too,...

in other way i can afford a regex type with codes matching, but at my first
test, the runtime excceded 100ms,
and there is much loss of memory in that case especially at adding timeout
execution option.

otherwise, the target type can be changed on inheriting from scalar
directly and produce money : amount
plus the currency codes or the culture language as a tuple or complexe
structure,

and it will be much usefull for c# users, especially if we afford
decimal(19;4) (the equivalent money on sql database too)

i waited a comment on that, i implemented just the basis, that will be much
near to 4217 ref

what you suggest?

Le lun. 15 mar. 2021 à 22:23, PascalSenn ***@***.***> a
écrit :

michaelstaib:
We are closing this PR since its stale.

Issue #3261, opened 2021-03-12
Add USCurrency scalar


wonbyte:
Add `USCurrency` scalar to `Hotchocolate.Types.Scalars`

- https://github.com/ChilliCream/hotchocolate/pull/3279

Addresses: #2927

wonbyte:
Closing in favor of LocalCurrency scalar.

PR #3279, opened 2021-03-14
Add USCurrency scalar

wonbyte:
Add `USCurrency` to `Hotchocolate.Types.Scalars`

Addresses #3261


wonbyte:
after talking with @PascalSenn it makes more sense to have a `LocalCurrency` instead.

PR #3288, opened 2021-03-15
Improved Rgb Scalar to allow percentage

PascalSenn:
Replaces #3260 as the fork was deleted


src/HotChocolate/Core/src/Types.Scalars/RgbType.cs
```diff
@@ -13,8 +13,7 @@ namespace HotChocolate.Types.Scalars
     public class RgbType : RegexType
     {
         private const string _validationPattern =
-            "^rgb\\(\\s*(-?\\d+|-?\\d*\\.\\d+(?=%))(%?)\\s*,\\s*(-?\\d+|-?\\d*\\.\\d+(?=%))("+
-            "\\2)\\s*,\\s*(-?\\d+|-?\\d*\\.\\d+(?=%))(\\2)\\s*\\)$";
+            "((?:rgba?)\\((?:\\d+%?(?:,|\\s)+){2,3}[\\s\\/]*[\\d\\.]+%?\\))";
```
PascalSenn, line 16: @wonbyte  wouldn't that make the Rgb scalar equivalent with the Rgba scalar?
I mean, when we have a Rgba a scalar these values should probably not be allowed in the Rgb scalar. Or did i understand something wrong? :)

PR #3294, opened 2021-03-16
Update migrate-from-10-to-11.md

kachniarzk:
Selection middleware (HotChocolate.Types.Selections) was replaced by projection middleware (HotChocolate.Data)

Recently my team was migrating HotChocolate to v11. The NuGet package HotChocolate.Types.Selections, that we were using, was only available in v10. After a long investigation, we found how to migrate properly, but it would be really helpful if the guide would contain information about projection middleware.


PascalSenn:
@kachniarzk  That's a good addition. Thanks for the contribution!

michaelstaib (approved):

PascalSenn (approved):

Issue #3263, opened 2021-03-12
Add UtcOffset scalar


wonbyte:
Add `UtcOffset` scalar type to `Hotchocolate.Types`

- https://github.com/ChilliCream/hotchocolate/pull/3319

Addresses: #2927

wonbyte:
closing for https://github.com/ChilliCream/hotchocolate/pull/3362

PR #3319, opened 2021-03-20
Add UtcOffset scalar

wonbyte:
Add `UtcOffset` scalar to `Hotchocolate.Types`

Addresses #3263


src/HotChocolate/Core/global.json
```diff
@@ -0,0 +1,5 @@
+{
```
PascalSenn, line 1: I dont think this file is needed :)

src/HotChocolate/Core/src/Types.Scalars/EmailAddressType.cs
```diff
@@ -1,6 +1,5 @@
 using System.Text.RegularExpressions;
 using HotChocolate.Language;
-using HotChocolate.Types.Scalars;
```
PascalSenn, line 3: Do these still compile without `HotChocoalte.Types.Scalars`? I mean, the resources are still in this package right?

src/HotChocolate/Core/src/Types.Scalars/ThrowHelper.cs
```diff
@@ -549,23 +547,23 @@ public static SerializationException RgbaType_ParseLiteral_IsInvalid(IType type)
...
         }
 
-        public static SerializationException UnsignedIntType_ParseValue_IsNotUnsigned(IType type)
+        public static SerializationException UnsignedIntType_ParseLiteral_IsNotUnsigned(IType type)
         {
             return new SerializationException(
                 ErrorBuilder.New()
-                    .SetMessage(ScalarResources.RgbType_IsInvalid_ParseValue)
-                    .SetCode(ErrorCodes.Scalars.InvalidRuntimeType)
-                    .SetExtension("actualType", WellKnownScalarTypes.Rgb)
```
PascalSenn, line 558: Good catch!

src/HotChocolate/Core/src/Types.Scalars/UtcOffsetType.cs
```diff
@@ -0,0 +1,188 @@
...
+                {new TimeSpan(8,0,0), "+08:00"},
+                {new TimeSpan(8,45,0), "+08:45"},
+                {new TimeSpan(9,0,0), "+09:00"},
+                {new TimeSpan(9,30,0), "+09:30"},
+                {new TimeSpan(10,0,0), "+10:00"},
+                {new TimeSpan(10,30,0), "+10:30"},
+                {new TimeSpan(11,0,0), "+11:00"},
+                {new TimeSpan(12,0,0), "+12:00"},
+                {new TimeSpan(12,45,0), "+12:45"},
+                {new TimeSpan(13,0,0), "+013:00"},
```
PascalSenn, line 51: ;)

src/HotChocolate/Core/src/Types.Scalars/UtcOffsetType.cs
```diff
@@ -0,0 +1,188 @@
...
+                {new TimeSpan(14,0,0), "+14:00"},
+            };
+
+        private static readonly Dictionary<string, TimeSpan> _offsetToTimeSpan =
```
PascalSenn, line 55: Can we map these from the _timeSpanToOffset dictionary?
something like  
```csharp
  var dict =  _offsetToTimeSpan
        .Reverse()
        .ToDictionary(x => x.Value, x=> x.Key);

    //Then add the missing things
    dict["+00:00"] = ...
    dict["-00:00"] = ...
```

src/HotChocolate/Core/src/Types.Scalars/UtcOffsetType.cs
```diff
@@ -0,0 +1,188 @@
...
+using HotChocolate.Language;
+
+namespace HotChocolate.Types
+{
+    /// <summary>
+    /// The `UtcOffset` scalar type represents a value of format ±hh:mm.
+    /// </summary>
+    public class UtcOffsetType : ScalarType<TimeSpan, StringValueNode>
+    {
+        private static readonly Dictionary<TimeSpan, string> _timeSpanToOffset =
```
PascalSenn, line 12: You may want to create a nested class to hold this things. It's just nice to work on the dictionaries when you have a constructor to do stuff :) 
 
At the end of the class:

```csharp
private class OffsetLookup 
{
     private readonly Dictionary<TimeSpan, string> _timeSpanToOffset;
     ....
     public OffsetLookup() 
     {
          //...fill the properties  
     }
     
     public bool TrySerialize(TimeSpan value, [NotNullWhen(true)] out string? result) {..}
     public bool TryDeserialize(string value, [NotNullWhen(true)] out TimeSpan? result) {..}
    
     ....
     public static readonly OffsetLookup = new OffsetLookup(); 
}
```

With this you can then just do 

```csharp
 public override IValueNode ParseResult(object? resultValue)
 {
     return resultValue switch
     {
          null => NullValueNode.Default,
          string s when OffsetLookup.TryParse(s, out TimeSpan? timespan) => ParseValue(timespan),
          TimeSpan ts => ParseValue(ts),
          _ => throw ThrowHelper.UtcOffset_ParseValue_IsInvalid(this)
      };
}
```

PascalSenn (changes requested):
Looks good ! :) 
Just a couple of small things

src/HotChocolate/Core/src/Types.Scalars/UtcOffsetType.cs
wonbyte: The only issue is what about `+00:00` vs. `-00:00`? When the `TimeSpan` is the key it's just 0 which we can't set twice..

PR #3350, opened 2021-03-24
Update operations.md

hallizh:
Just a proposal so that people are not in the dark when it comes to subscriptions. Just a bit more helpful than the current placeholder but will hopefully not stick around for long.


hallizh:
I just got word that the documentation is a WIP and is almost ready, closing this.

PR #3415, opened 2021-04-01
Fixes the error that caused the side not being scrollable

fredericbirke:
Fixes the error that caused the side not being scrollable by dragging the scrollbar on chrome within sites that have a sidebar with a toc


website/package.json
```diff
@@ -61,7 +61,8 @@
     "react-redux": "^7.2.1",
     "react-responsive-carousel": "^3.2.11",
     "react-share": "^4.2.1",
-    "styled-components": "^5.2.0"
+    "redux-devtools-extension": "^2.13.9",
```
rstaib, line 64: This is a dev dependency.

website/src/components/articles/article-sections.tsx
```diff
@@ -37,17 +37,23 @@ export const ArticleSections: FunctionComponent<ArticleSectionsProperties> = ({
...
   }, [data]);
 
   useEffect(() => {
-    setActiveHeadingId(
-      headings.find((id) => yScrollPosition >= id.position)?.id
+    const activeHeading = headings.find((id) => yScrollPosition >= id.position)
+      ?.id;
+    window.history.pushState(
+      undefined,
+      activeHeading || "ChilliCream Docs",
```
rstaib, line 53: Let's use `??`.

website/src/components/doc-page/doc-page-navigation.tsx
```diff
@@ -256,16 +256,17 @@ export const DocPageNavigation: FunctionComponent<DocPageNavigationProperties> =
...
+                        path: item!.path!,
+                        title: item!.title!,
+                      }))
                   : undefined,
               })),
-            `/docs/${activeProduct!.path!}${activeVersion?.path?.length && activeVersion.path.length > 0
-              ? "/" + activeVersion.path!
-              : ""
+            `/docs/${activeProduct!.path!}${
+              activeVersion?.path?.length && activeVersion.path.length > 0
```
rstaib, line 267: Use `!!activeVersion?.path?.length`instead of `activeVersion?.path?.length && activeVersion.path.length > 0`.

rstaib: I now that you just reformatted that, but when touching then also improve 😄

website/src/components/misc/page-top.tsx
```diff
@@ -9,32 +9,22 @@ export const PageTop: FunctionComponent<{ onTopScroll: () => void }> = ({
...
+    (state) => state.common.yScrollPosition <= 60
   );
 
   return (
-    <JumpToTop
-      className={showButton ? "show" : ""}
-      ref={ref}
-      onClick={onTopScroll}
-    >
+    <JumpToTop hideButton={hideButton} ref={ref} onClick={onTopScroll}>
```
rstaib, line 17: This is less performant than using a class.

website/src/docs/hotchocolate/api-reference/extending-filtering.md
```diff
@@ -284,8 +284,7 @@ A little simplified this is what happens during visitation:
 ```graphql
 {
   users(
-    where: # instance[0] = x # Create SCOPE 1 with parameter x of type User
-    # level[0] = []
+    where: # level[0] = [] # instance[0] = x # Create SCOPE 1 with parameter x of type User
```
rstaib, line 287: I guess each `#` represents a new line.

rstaib (changes requested):
Look's good; Just some minor things I would like to be seen fixed 😸

website/src/components/doc-page/doc-page-navigation.tsx
fredericbirke: I think these changes come from running prettiert :D?

website/src/docs/hotchocolate/api-reference/extending-filtering.md
fredericbirke: From prettier as well, shall I revert it manually? Maybe we should skip .md files with prettier?

website/src/components/misc/page-top.tsx
fredericbirke: The class was causing the chrome scroll bar to stop. I think it is some kind of issue with chrome, I found older issues with similar problems in the chrome bugtracker. It works fine in all other browsers, but not in chromium. Without using the class it works fine. But also not with all css properties :D Some still reset the scrollbar state. That is also the reason for:
```
height: 0;
overflow: hidden;
box-shadow: none;
```

because setting `display: none` would cause the scrollbar to bug out again.

PR #3494, opened 2021-04-13
Document custom request size limits with Multipart

tobias-tengler:
I added a note about request size limits to the documentation.

I also changed one of the existing multipart error messages to include the exception detailing why the parsing of the multipart form failed. This should help debugging issues related to request sizes and badly formed multipart requests.

Relates to #3473


michaelstaib (dismissed):

tobias-tengler:
Oh, I think I forgot to run tests again, after including the Exception. I will fix it!

tobias-tengler:
@michaelstaib I thought adding the exception to the ErrorBuilder would include it in the output, but it isn't inlcuded:
image
How would I include the original exception? I think this would give meaningful insights into how to fix the issue (at least in development).

michaelstaib:
Adding the exception will make the exception available to internal diagnostics. This means that if you attach a logger or intercept the error you will be able to catch it. If you want to expose the internal error message you should also add an extension field to the error:

```csharp
ErrorBuilder.New()
   ....
   .SetExtension("underlyingError", ex.Message)
```

Just add this additionally.

tobias-tengler:
Added the extension field and also added a test that showcases what happens if a file is uploaded that exceeds limits.

michaelstaib:
@tobias-tengler I have given you write access to the repo, you can now add your own branches to the main repo which makes contributing easier.

The branch names are structured like follows: `{3 letters representing your name}/{branch name}`. 

example `mst/some-feature`.

michaelstaib (approved):

PR #3548, opened 2021-04-19
Document Pagination

tobias-tengler:


tobias-tengler:
@michaelstaib What is the current best practice for implementing custom cursor based pagination? Let's say I want to add a `sortBy` argument to my resolver.
Would the following be the best way to go about it?
```csharp
descriptor
    .Field("users")
    .UsePaging<UserType>()
    .Argument("sortBy", a => a.Type<StringType>())
    .Resolve(context =>
    {
        var first = context.ArgumentValue<int?>("first");
        // ...
        var sortBy = context.ArgumentValue<string?>("sortBy");

        IEnumerable<User> users = null; // get users using the above arguments

        var edges = users.Select(user => new Edge<User>(user, user.Id)).ToList();
        var pageInfo = new ConnectionPageInfo(false, false, null, null);

        var connection = new Connection<User>(edges, pageInfo, ct => new ValueTask<int>(0));

        return connection;
    });
```

From my understanding the `CursorPagingHandler` can be used to implement custom slicing behavior for data returned by my resolver. I would use this, if I need to implement a re-useable pagination behavior for a set of specific return types, right?
Should the `CursorPagingHandler` be covered by the pagination documentation?

tobias-tengler:
I'm probably going to get rid of the first part, where I explained the difference between offset and cursor based pagination. It's probably too opinionated and out of scope for this project...

tobias-tengler:
TODOS:
- [ ] Custom edges code example
- [ ] Schema-first code examples
- [ ] SetPagingOptions

website/src/docs/hotchocolate/fetching-data/pagination.md
```diff
@@ -2,12 +2,575 @@
...
+
+Using `WHERE` instead of `OFFSET` is great, since now we can leverage the index of the `Id` field and the database does not have to compute an _offset_.
+
+For this to work though, our _cursor_ needs to be **unique** and **sequential**. Most of the time the _Id_ field will be the best fit.
+
+But what if we need to sort by a field that does not have the aforementioned properties? We can simply combine the field with another field, which has the needed properties (like `Id`), to form a _cursor_.
+
+Let's look at another example: We want to paginate over the users sorted by their birthday.
+
+After receiving the first page, we create a combined _cursor_, like `"1435+2020-12-31"` (`Id` + `Birthday`), of the next entry. To receive the second page, we convert the _cursor_ to its original values (`Id` + `Birthday`) and use them in our query:
```
tobias-tengler, line 72: Note to self: Revisit this section

tobias-tengler:
@michaelstaib I've commented out the custom edges part for now. I think we should get this merged asap to complete the "Fetching data" chapter. We can revisit it for the custom edges part later (I'm really interested in it, but I think it's better not to stall this PR for it). 
I've also tried to implement the examples with schema-first, but I had no success. I labeled them as unsupported for now.

michaelstaib (approved):

PR #3618, opened 2021-04-29
docs: remove obsolete function refrence

steveoh:
replace with non obsolete function name


michaelstaib:
https://github.com/ChilliCream/hotchocolate/pull/3584

michaelstaib:
This is the V10 docs where it is not obsolete. The new docs are in 3584

steveoh:
Ok, then image has tricked me more than once. How about making it more obvious than font weight which docs we're looking at.

PR #3636, opened 2021-04-30
Track changes to public API

### LINKED ISSUE #3532: Track changes to Public API
We could add a Tool like the [Roslyn Public Api Analyzer](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) to the project to check for breaking changes.

This would have a number of benefits:
- Breaking changes don't need to be remembered, we can now simply diff a file between git tags.
- Breaking changes are less likely to ship per accident
- If there is a breaking change, we can simply search the repo for the old usage and switch it over. This should prevent examples and documentation from getting out-of-date.

I haven't looked deeply into the topic yet, I just wanted to propose this. Maybe there are also other tools that are better, but this tool is for example used by the ASP.NET Core team.

tobias-tengler:
This adds the [PublicApiAnalyzers](https://github.com/dotnet/roslyn-analyzers/tree/main/src/PublicApiAnalyzers) roslyn analyzer to track changes to the public API surface of HC. I was inspired to include this after I've stumbled upon it in the [ASP.NET Core repository](https://github.com/dotnet/aspnetcore/blob/52eff90fbcfca39b7eb58baad597df6a99a542b0/docs/APIBaselines.md#publicapishippedtxt).

## How?

What it does is, it includes two new files in every project (except Tests, Resources and manually excluded projects):
- `PublicAPI.Shipped.txt`
- `PublicAPI.Unshipped.txt`

These files contain a text-based representation of HC's API surface. If the actual code changes, but it hasn't changed in these files the analyzer gives a warning (or error if configured). I have already populated the `PublicAPI.Shipped.txt` files using the Code-Fix of the analyzer.

Let's see what happens if I add a new public property to a public class:

image
If I execute the Code-Fix, Visual Studio would write the following to `PublicAPI.Unshipped.txt`:
```
GreenDonut.Batch<TKey, TValue>.NewProperty.get -> string?
GreenDonut.Batch<TKey, TValue>.NewProperty.set -> void
```
Checkout [this document](https://github.com/dotnet/aspnetcore/blob/52eff90fbcfca39b7eb58baad597df6a99a542b0/docs/APIBaselines.md) for the update/delete case.

After a new major version has been published, one person manually runs the `mark-shipped.ps1` script, pulls all `PublicAPI.Unshipped.txt` changes into `PublicAPI.Shipped.txt` and checks it in, therefor creating a new baseline.

## Why?

- Having warnings / errors in your IDE helps prevent possible breaking changes early on
- CI could error out for PRs with undocumented API changes - Maintainers can quickly detect changes by looking at these files
- A text-based definition of the public API surface can be easily used to visualize differences between versions, think `git diff`
- Functionality that should really be internal, but is public by accident, can be spotted more easily
- Seeing these changes more clearly in one place can help keep documentation up-to-date

## Potential problems

- The Code-Fix currently only works in IDEs that support Roslyn Analyzers, e.g. Visual Studio (VS Code is not supported)
  EDIT: I'm working on [this issue](https://github.com/OmniSharp/omnisharp-roslyn/issues/1592) to get support going for VS Code via Omnisharp)
- Keeping these files up-to-date (even with Code-Fixes) adds some overhead

Closes #3532


tobias-tengler:
I've added some more scripts that could be useful.

## display-unshipped-api.ps1
When called this will output the contents of all `PublicAPI.Unshipped.txt` files throughout the project:

image
This could be useful to see all changes once more, before executing `mark-api-shipped.ps1`.

## mark-api-shipped.ps1
This transfers all changes in the `PublicAPI.Unshipped.txt` to the `PublicAPI.Shipped.txt` files. It also takes care of removing lines marked with `*REMOVE*` (removals of APIs).

## diff-shipped-api.ps1
This shows all changes of `PublicAPI.Shipped.txt` files between git refs. 
I could therefore execute `diff-shipped-api.ps1 -from 11.0.0 -to 12.0.0` to see all APIs that changed between those two tags:

image
This could be used to generate a document containing all API changes between major version.

Note: I would like to get rid of the `git diff` header as well, but I haven't been able to yet.

tobias-tengler:
I've made it so that locally during development only warnings are shown, but when the nuke build is run, i.e. CI, the warnings are treated as errors.

With this you can still play around locally and build everything without issues, but once it comes to merging you're forced to freeze any changes to the public API. I think this setup provides the most flexibility whilst still tracking the public API on the main branch.

tobias-tengler:
This PR is ready from my side! 🚀

michaelstaib (approved):

PR #3651, opened 2021-05-01
Added error for pagination in projections

PascalSenn:
Added error message for pagination in projections


PascalSenn:
@michaelstaib  should we merge this one? nested pagination is currently not supported but might be in a future version

PR #3675, opened 2021-05-07
Neo4j Data Integration Initial

arif-hanif:


PascalSenn:
I reviewed the PR locally and did not yet leave a review on the PR itself. Looks good @arif-hanif, well done!

# Formatting and code style #3700 
The PR #3700 aligns the code style more with the core projects. I mainly copied things around and did some styling fixes.
Member ordering, indentation, if without braces, and so on.

# Language
## Properties #3701
I checked the language package and saw that there have been many methods like `GetXYZ` that expose a private field `XYZ`. In .NET we usually use properties instead of fields and getters.
At the moment, there are still a few parts that only expose a part of their members (because they are only used internally).
I think we might want to just expose everything that defines the language node 

As an example, in the following class we would also expose `_value` as `Value`:
```csharp
    public class KeyValueMapEntry : Expression
    {
        private readonly Expression _value;

        public KeyValueMapEntry(string key, Expression value)
        {
            Key = key;
            _value = value;
        }

        public override ClauseKind Kind => ClauseKind.KeyValueMapEntry;

        public string Key { get; }
```


## Internal
The language part of the API is quite big and I think after we released the Neo4J package, we might want to be able to change a few bits and pieces in there. (e.g. Visitor Logic, what is exposed and what not, etc.)

I would strongly suggest to make everything of the language `internal` for now, so that we can iterate on it a bit more. 

## AsString -> Print
I was not sure about that but still included it in #3701. In the core, we either override `ToString()` or add a method `Print()` to visualize the value or types. 
I see that in Neo4J this is also used to build the Cypher at the end, so I am not sure if `AsString()` should stay. Happy to revert if you guys do not agree.

## XML documentation
I would love to see XML documentation on these literals, but this can wait. I think, if we make the Language part internal we can still work on the XML docs before we open it up to the public. 

## Immutability 
At the moment, most of the language is immutable. But a few things, like the `CompoundCondition` has local mutable state. I think we should define if members of the language should be immutable or mutable.
My vote goes to the immutability of everything 

## Add != With
A method of an immutable API is prefixed with `With`. So instead of `Property` we should use `WithProperty` and so on.

# Dead Code and Comments
As this pr has been growing over a few months now, I found many places where there are `NotImplementedException`s, not finished XML Documentation or comments. This, is something we should cleanup before the release.

# Nullability
There are a lot of nullability issues all over the place. This shows that something might not be correctly abstracted yet. But we will see as we fix them :) 

# Ensure.cs
Have a look at https://github.com/ChilliCream/hotchocolate/blob/neo4j/src/HotChocolate/Neo4J/src/Data/Utilities/Ensure.cs

`Ensure` provides functionality to do argument validation like
```csharp
public static CompoundCondition Create(Condition left, Operator op, Condition right)
{
    Ensure.IsTrue(
        _validOperations.Contains(op),
        "Operator is not a valid operator for a compound condition.");
    Ensure.IsNotNull(left, "Left hand side condition is required.");
    Ensure.IsNotNull(op, "Operator is required.");
    Ensure.IsNotNull(right, "Right hand side condition is required.");
    return new CompoundCondition(op)
        .Add(op, left)
        .Add(op, right);
}
```

Personally, I am open for something like this, but I guess it's kind of an all or nothing situation. 
Either we adopt it in the core and start using it everywhere, or we leave it out and validate only the important parts. 

("only the important parts": e.g. as long as the language is internal we do not have to check for IsNotNull)

# Attributes
Do we need all the attributes defined in https://github.com/ChilliCream/hotchocolate/tree/neo4j/src/HotChocolate/Neo4J/src/Data/Attributes ? 
Specifically `Neo4JIgnoreAttribute`, `Neo4JNodeAttribute`, `Neo4JNodeIdAtribute` and `Neo4JDateTimeAttribute`.
I would also like to consider removing the Neo4J prefix on the once we actually need.

If we need them to translate from query to cypher we should also provide extension methods on `IObjectFieldDescriptor` et al. 

# Memory Leak
I think we also have a memory leak in the CypherWriter.
We do pool the string builders here and return it to the pool on dispose
https://github.com/ChilliCream/hotchocolate/pull/3675/files#diff-82c21ad36ec370210f28ed7fce1eb611a549823e0743c81739ef59ae9bd20f21R9-R17

That is as such not an issue. The problem is that we use the CypherWriter in the CypherVisitor.
The CypherVisitor does not implement IDisposable and never disposes the CypherWriter it instantiates.
Meaning, the CypherWriter is never returned to the pool

# A few other things
I have a list of a few other things that I would like to point out or ask why they are like they are.
But I guess this makes more sense if I comment them on the PR on the specific code passages.

# Next Steps
It would make sense to first merge #3700 and #3701 before we do any further changes, as they would possibly conflict with these PRs
Then I will have a look at the PR and point out the parts that are on my list. 
We can then divide the work up to polish this PR for a preview release 

# Open Tasks
- [ ] Code styling and review of `*.Tests*` 
- [ ] Make language internal 
- [ ] Language immutability
- [ ] Define strategy for Ensure
- [ ] Fix Memory Leak 
- [ ] Revisit Attributes 
- [ ] Fix nullability issues
- [ ] Remove Dead Code & NotImplementedExceptions
- [ ] XML Documentation for the public parts
- [ ] Reenable Public API analyzer
- [ ] Fix comments on the PR

src/HotChocolate/Neo4J/src/Data/Language/Skip.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            cypherVisitor.Enter(this);
+            _skipAmount.Visit(cypherVisitor);
+            cypherVisitor.Leave(this);
+        }
+
+        public static Skip Create(int value)
+        {
+            return new(new IntegerLiteral(value));
+        }
```
PascalSenn, line 29: Not in use

src/HotChocolate/Neo4J/src/Data/Language/SortItem.cs
```diff
@@ -0,0 +1,37 @@
...
+        public SortItem(Expression expression, SortDirection? direction)
+        {
+            _expression = expression;
+            _direction = direction;
+        }
+
+        public override ClauseKind Kind => ClauseKind.SortItem;
+
+        public SortItem Ascending() => new(_expression, SortDirection.Ascending);
+        public SortItem Descending() => new(_expression, SortDirection.Descending);
```
PascalSenn, line 17: With*

src/HotChocolate/Neo4J/src/Data/Language/Statement.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace HotChocolate.Data.Neo4J.Language
+{
+    public abstract class Statement : Visitable
+    {
+        public override ClauseKind Kind => ClauseKind.Statement;
+
+        public abstract Dictionary<string, object> GetParameters();
+
+        public abstract List<string> GetParameterNames();
+
```
PascalSenn, line 12: Seems not to be in use

src/HotChocolate/Neo4J/src/Data/Language/StatementBuilder.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public StatementBuilder Return(params Expression[]? expressions)
+        {
+            _return = new Return(new ExpressionList(expressions));
+            return this;
+        }
+
+        public StatementBuilder Where(Condition condition)
+        {
+            //_where = new Where(condition);
```
PascalSenn, line 62: This seems odd

src/HotChocolate/Neo4J/src/Data/Language/SymbolicName.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public static SymbolicName Of(string name)
+        {
+            Ensure.HasText(name, "Name must not be empty.");
+            return new SymbolicName(name);
+        }
+
+        public static SymbolicName Unresolved() => new(null);
+
+        public static SymbolicName Unsafe(string name)
```
PascalSenn, line 32: Why is this unsafe?

src/HotChocolate/Neo4J/src/Data/Language/SymbolicName.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public string Value { get; }
+
+        public static SymbolicName Of(string name)
+        {
+            Ensure.HasText(name, "Name must not be empty.");
+            return new SymbolicName(name);
+        }
+
+        public static SymbolicName Unresolved() => new(null);
```
PascalSenn, line 30: Nullability

src/HotChocolate/Neo4J/src/Data/Language/Visitor/CypherVisitor.VisitationMap.cs
```diff
@@ -0,0 +1,231 @@
+using System.Linq;
+
+namespace HotChocolate.Data.Neo4J.Language
+{
+    public partial class CypherVisitor
```
PascalSenn, line 5: At some point we can align this with the core and visit from the out side in. 
Then you can just instantiate a visitor and override what you need

src/HotChocolate/Neo4J/src/Data/Language/Visitor/CypherVisitor.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace HotChocolate.Data.Neo4J.Language
+{
+    public partial class CypherVisitor
+    {
+        private readonly LinkedList<IVisitable> _currentVisitedElements = new();
+
+        /// <summary>
+        /// Target for all rendered parts of the cypher statement.
+        /// </summary>
+        private readonly CypherWriter _writer = new();
```
PascalSenn, line 13: This is a memory leak. The writer implements IDispoable and gets StringBuilders from the pool. But the cypher visitor is not Disposable and the builder will never be returned back to the pool

src/HotChocolate/Neo4J/src/Data/Language/Visitor/CypherVisitor.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Keeps track of unresolved symbolic names.
+        /// </summary>
+        private readonly Dictionary<SymbolicName, string> _resolvedSymbolicNames = new();
+
+        /// <summary>
+        /// The current level in the tree of cypher elements.
+        /// </summary>
+        private int _currentLevel;
```
PascalSenn, line 45: What you could do to avoid the `++`, `+1` etc. is to introduce a Scope.
Like in the filter visitor. 
On a scope you have everything that is related to this scope

src/HotChocolate/Neo4J/src/Data/Language/Visitor/CypherWriter.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </summary>
+    internal sealed class CypherWriter : IDisposable
+    {
+        private bool _isDisposed;
+
+        private static readonly ObjectPoolProvider _objectPoolProvider =
+            new DefaultObjectPoolProvider();
+
+        private static readonly ObjectPool<StringBuilder> _stringBuilderPool =
+            _objectPoolProvider.CreateStringBuilderPool();
```
PascalSenn, line 18: Make sure to always disposethe CypherWriter

src/HotChocolate/Neo4J/src/Data/Language/Where.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// WHERE adds constraints to the patterns in a MATCH or OPTIONAL MATCH clause or filters the
+    /// results of a WITH clause.
+    /// <a href="https://s3.amazonaws.com/artifacts.opencypher.org/railroad/Match.html#Where" >
+    /// Where
+    /// </a>
+    /// </summary>
+    public class Where : Visitable
+    {
+        private readonly Exists _exists;
```
PascalSenn, line 12: This can be null in some cases

src/HotChocolate/Neo4J/src/Data/Paging/Neo4JOffsetPagingProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+                Neo4JExecutable<TEntity> f = CreatePagingContainer(source);
+                return ResolveAsync(context, f, arguments);
+            }
+
+            private Neo4JExecutable<TEntity> CreatePagingContainer(object source)
+            {
+                return source switch
+                {
+                    Neo4JExecutable<TEntity> nfe =>
+                        CreatePagingContainer(nfe.Pipeline()),
```
PascalSenn, line 74: Doesn't this always lead to a exception?

src/HotChocolate/Neo4J/src/Data/Paging/Neo4JOffsetPagingProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+
+                queryable = queryable.WithLimit(arguments.Take + 1);
+
+                IList items = await queryable
+                    .ToListAsync(context.RequestAborted)
+                    .ConfigureAwait(false);
+
+
+                return new CollectionSegment((IReadOnlyCollection<object>)items, null, null);
```
PascalSenn, line 96: 2nd and 3third attribute cannot be null

src/HotChocolate/Neo4J/src/Data/Paging/UseNeo4JOffsetPagingAttribute.cs
```diff
@@ -0,0 +1,82 @@
...
+using HotChocolate.Types.Pagination;
+
+namespace HotChocolate.Data.Neo4J.Paging
+{
+    /// <summary>
+    /// This attribute adds the offset paging middleware for Neo4j to the annotated method or
+    /// property.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Method)]
+    public class UseNeo4JOffsetPagingAttribute : DescriptorAttribute
```
PascalSenn, line 14: We will try to get rid of specific attribute before the release

src/HotChocolate/Neo4J/src/Data/Projections/Handlers/Neo4JProjectionFieldHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+        public override bool CanHandle(ISelection selection) =>
+            selection.SelectionSet is not null;
+
+        /// <inheritdoc/>
+        public override bool TryHandleEnter(
+            Neo4JProjectionVisitorContext context,
+            ISelection selection,
+            [NotNullWhen(true)] out ISelectionVisitorAction? action)
+        {
+            ++context.CurrentLevel;
```
PascalSenn, line 25: context.AddScope()

src/HotChocolate/Neo4J/src/Data/Projections/Handlers/Neo4JProjectionFieldHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+                            .Enqueue(projections);
+                        break;
+                    case 1:
+                        context.Projections.Push(field);
+                        context.Projections.Push(projections);
+                        break;
+                }
+            }
+
+            --context.CurrentLevel;
```
PascalSenn, line 80: context.PopScope()

src/HotChocolate/Neo4J/src/Data/Projections/Neo4JProjectionProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+                        context.Result = Array.Empty<TEntityType>();
+                        foreach (IError error in visitorContext.Errors)
+                        {
+                            context.ReportError(error.WithPath(context.Path));
+                        }
+                    }
+                    else
+                    {
+                        context.LocalContextData =
+                            context.LocalContextData.SetItem("ProjectionDefinition", projections);
```
PascalSenn, line 56: This string should be a public const somewhere

src/HotChocolate/Neo4J/src/Data/Projections/Neo4JProjectionVisitorContext.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        /// <inheritdoc/>
+        public Neo4JProjectionVisitorContext(
+            IResolverContext context,
+            IOutputType initialType)
+            : base(context, initialType, new Neo4JProjectionScope())
+        {
+        }
+
+        public int CurrentLevel { get; set; }
```
PascalSenn, line 21: Scopes

src/HotChocolate/Neo4J/src/Data/Sorting/Neo4JSortDefinition.cs
```diff
@@ -0,0 +1,18 @@
+﻿using HotChocolate.Data.Neo4J.Language;
+
+namespace HotChocolate.Data.Neo4J.Sorting
+{
+    public class Neo4JSortDefinition
+    {
+        public Neo4JSortDefinition(string field, SortDirection direction)
```
PascalSenn, line 7: Could be immutable

src/HotChocolate/Neo4J/src/Data/Sorting/Neo4JSortProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+                        context.Result = Array.Empty<TEntityType>();
+                        foreach (IError error in visitorContext.Errors)
+                        {
+                            context.ReportError(error.WithPath(context.Path));
+                        }
+                    }
+                    else
+                    {
+                        context.LocalContextData =
+                            context.LocalContextData.SetItem("Sorting", sorts);
```
PascalSenn, line 65: public const

PascalSenn (changes requested):

src/HotChocolate/Neo4J/src/Data/Attributes/Neo4JIgnoreAttribute.cs
```diff
@@ -0,0 +1,23 @@
+using System.Reflection;
+using HotChocolate.Types;
+using HotChocolate.Types.Descriptors;
+
+namespace HotChocolate.Data.Neo4J
+{
+    /// <summary>
+    /// If this is used on a field it will be ignored.
+    /// </summary>
+    public class Neo4JIgnoreAttribute
```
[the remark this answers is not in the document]
arif-hanif: long term yes, but can pull out for now

src/HotChocolate/Neo4J/src/Data/Attributes/Neo4JNodeAttribute.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using HotChocolate.Types;
+using HotChocolate.Types.Descriptors;
+
+namespace HotChocolate.Data.Neo4J
+{
+    public class Neo4JNodeAttribute : ObjectTypeDescriptorAttribute
```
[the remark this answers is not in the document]
arif-hanif: long term yes, but can pull out for now

src/HotChocolate/Neo4J/src/Data/Attributes/Neo4JNodeIdAttribute.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace HotChocolate.Data.Neo4J
+{
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    public class Neo4JNodeIdAttribute : Attribute
```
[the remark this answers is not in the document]
arif-hanif: long term yes, but can pull out for now

src/HotChocolate/Neo4J/src/Data/Attributes/Neo4jDateTimeAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+using HotChocolate.Types;
+using HotChocolate.Types.Descriptors;
+
+namespace HotChocolate.Data.Neo4J
+{
+    /// <summary>
+    /// If this is used on a <see cref="System.DateTime"/> property - it will be serialized as a
+    /// Neo4j Date object rather than a string.
+    /// </summary>
+    public class Neo4JDateTimeAttribute
```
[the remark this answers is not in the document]
arif-hanif: long term yes, but can pull out for now

src/HotChocolate/Neo4J/src/Data/Filtering/Handlers/List/Neo4JListAnyOperationHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+                context.RuntimeTypes.Peek().TypeArguments is { Count: > 0 } &&
+                parsedValue is bool parsedBool &&
+                context.Scopes.Peek() is Neo4JFilterScope)
+            {
+                if (parsedBool)
+                {
+                    return new CompoundCondition(Operator.And);
+                }
+
+                return new CompoundCondition(Operator.And);
```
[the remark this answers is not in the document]
arif-hanif: not complete, need to finish

PR #3693, opened 2021-05-15
Add Json Scalar

### LINKED ISSUE #3661: Add Json Scalar
Add `JsonType`  to `Hotchocolate.Types`

Addresses #2927

wonbyte:
Add `JsonType` to `Hotchocolate.Types`

Closes #3661


benmccallum:
Nice work! Yet another scalar we have at AutoGuru we can possible get rid of :P But ours is just a string, doesn't check validity or anything. 

We also have a HTML one if that's something that makes sense to add as well. Bit of a trickier one though as we may need to think about encoding so that out-of-the-box it's safe to inject raw into the client web page.

src/HotChocolate/Core/src/Types.Scalars/JsonType.cs
```diff
@@ -0,0 +1,65 @@
+using System;
+using System.Text.Json;
+using HotChocolate.Language;
+using HotChocolate.Utilities;
+
+namespace HotChocolate.Types
+{
+    public class JsonType : ScalarType
```
benmccallum, line 8: Reminder to add this into the docs in the table that lists all these

wonbyte:
closing for https://github.com/ChilliCream/hotchocolate/pull/3713

PR #3739, opened 2021-05-26
Add information about last update to documentation

### LINKED ISSUE #3351: Embed commit date to chillicream website docs
No

I would like that the document section of chillicream.com is embedding commit date for the document, this is so users won't have to go over to github to see when the documentation was last updated. 

N/A

**Additional context**

Original version:
image

Mocked version:
image

tobias-tengler:
I experimented with the date being below the heading as described in the issue, but it looked off to me no matter where I placed it.
So I ended up putting it on the bottom of the page, similar to how Docusaurus does it:

image

Closes #3351

### DIFF
--- a/website/gatsby-node.js
+++ b/website/gatsby-node.js
@@ -129,7 +129,8 @@ exports.onCreateNode = async ({ node, actions, getNode }) => {
     value: filepath,
   });
 
-  let authorName, lastUpdated;
+  let authorName;
+  let lastUpdated;
 
   // we only run "git log" when building the production bundle
   // for development purposes we fallback to dummy values
@@ -227,7 +228,7 @@ function createDocPages(createPage, data) {
   });
 }
 
-async function getGitLog(filepath) {
+function getGitLog(filepath) {
   const logOptions = {
     file: filepath,
     n: 1,
@@ -237,5 +238,5 @@ async function getGitLog(filepath) {
     },
   };
 
-  return await git().log(logOptions);
+  return git().log(logOptions);
 }
--- a/website/gatsby-node.js
+++ b/website/gatsby-node.js
@@ -1,5 +1,7 @@
 const { createFilePath } = require("gatsby-source-filesystem");
-const path = require(`path`);
+const path = require("path");
+const git = require("simple-git/promise");
+const moment = require("moment");
 
 exports.createPages = async ({ actions, graphql, reporter }) => {
   const { createPage, createRedirect } = actions;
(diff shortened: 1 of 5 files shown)


website/gatsby-node.js
```diff
@@ -112,17 +114,49 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
...
+
+  const filepath = createFilePath({ node, getNode });
+
+  createNodeField({
+    name: `slug`,
+    node,
+    value: filepath,
+  });
+
+  let authorName, lastUpdated;
```
rstaib, line 132: One let per variable, please 😉

website/src/components/articles/article-content-footer.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { FC } from "react";
+import styled from "styled-components";
+
+interface Props {
```
rstaib, line 4: To be consistent lets call it `ArticleContentFooterProperties`.

website/src/components/articles/article-content-footer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  lastUpdated: string;
```
rstaib, line 5: Always readonly for new or touched interfaces, please. I will soon update the entire codebase so that readonly is always used.

website/src/components/articles/article-content-footer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  lastAuthorName: string;
+}
+
+export const ArticleContentFooter: FC<Props> = ({
```
rstaib, line 9: For consistency and readability purposes, I prefer FunctionComponent. I'm generally open to discuss that, but for now I would prefer consistency. Thank you.

website/gatsby-node.js
```diff
@@ -192,3 +226,16 @@ function createDocPages(createPage, data) {
...
+  const logOptions = {
+    file: filepath,
+    n: 1,
+    format: {
+      date: `%ai`,
+      authorName: `%an`,
+    },
+  };
+
+  return await git().log(logOptions);
```
rstaib, line 240: `async/await` is not needed in this case, because the promise is immediately returned. Just return the Promise. It's more efficient in doing so, because it prevents from generating an unnecessary state machine for this particular function.

website/src/components/articles/article-content-footer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  lastUpdated,
+  lastAuthorName,
+}) => {
+  return (
+    <Footer>
+      <Small>
```
rstaib, line 15: A link to the git history to this file would be probably nice. What do you think?

rstaib (changes requested):
Well done! Just a few things. By the way it's really cool that you help us on the website and documentation stuff. Very much appreciated. Thank you!

website/src/components/articles/article-content-footer.tsx
tobias-tengler: Made the change for consistency, but I personally prefer FC. In my opinion it makes the code more readable, since there's less clutter.

website/src/components/articles/article-content-footer.tsx
tobias-tengler: Made the change, but I don't see the point in using the longer name, if we're not exporting it. Also not a big fan of `Properties` over `Props`. But if that's a convention here I'll go with it :)

website/src/components/articles/article-content-footer.tsx
tobias-tengler: I can't follow. What exactly do you mean?

website/src/components/articles/article-content-footer.tsx
rstaib: We can discuss conventions of course. `Props` is fine with me, but then consistent. We can change later.

website/src/components/articles/article-content-footer.tsx
rstaib: We can discuss conventions of course. FC is fine with me, but then consistent. If you would like to change, then go ahead, but please do in a separate branch and for the entire codebase. Thanks!

website/src/components/articles/article-content-footer.tsx
rstaib: Sorry, I mean it would be also cool to have an additional link for every markdown document (perhaps placed next to the last document change date), which points directly to document's commit history. Does this make more sense 😄

rstaib (approved):
Thank you for being patient! Was pretty busy this weekend.

PR #3746, opened 2021-05-27
Detect and fix broken links

tobias-tengler:
I essentially moved the code of [gatsby-remark-check-links](https://github.com/trevorblades/gatsby-remark-check-links) into this repository and made some changes. The original plugin has some flaws that make it incompatible with our website and it is also no longer maintained by the original author.

### Changes

- Add plugin to discover headings and links inside of markdown files
- Add plugin to check for the validity of links between markdown pages
- The slug field is now the actual slug (before: `/hotchocolate/get-started/`, now: `/docs/hotchocolate/get-started`)
- Centralize slug generation (page creation now uses the generated slugs)
- Fix dead links in the documentation and blog posts


rstaib:
@tobias-tengler could you please solve the conflicts so I can start the review 😄

tobias-tengler:
Already in the process of doing so :wink:

website/plugins/detect-broken-links/index.js
```diff
@@ -0,0 +1,209 @@
...
+
+    if (cacheEntry) {
+      documentMap[cacheEntry.slug] = cacheEntry;
+
+      continue;
+    }
+
+    // don't continue if a page hasn't been visited yet
+    return markdownAST;
+  }
```
tobias-tengler, line 103: @rstaib I don't really like this part. Essentially it checks if all markdown files had their links/headings added to the cache:
- If not, exit
- If they have, continue
This means the loop checks every file for every file emitted... The original plugin was doing it like that and I couldn't find a more suitable spot to put the logic after all files have been checked.
There are additional events like `onPostBuild`, etc. but in them I would have to be able to access all of the files and get their cache entries (essentially what is happening in the loop, but only executed once). Any ideas?

tobias-tengler:
Okay I've noticed two issues with this just now. 
1. When running `yarn build` it works as expected. But with `yarn start` it doesn't since the `slug` is not present on the `fields`.
2. With `yarn start` it also struggles with `.md` files that aren't in the `docs.json` file. I think that's because I check for each file with an .md extension and not just the valid ones.

Will look into these problems tomorrow.

tobias-tengler:
Okay I can't get it to work with `yarn start`. This is due to the "lazy" nature of gatsby in development. Not all pages are created at the same time, causing issues when trying to validate links. The markdown plugin is also always invoked before a slug for the node is generated, therefor we don't know under which URL the current markdown document will be available. This is not an issue with `yarn build`.

I think it is fine like it is though. It will discover broken links in CI and when `yarn build` is invoked locally and that's better than no indication at all.

I also significantly improved the performance in comparison to my initial attempt, by splitting the gathering of links and validation of those links into two separate local plugins.

rstaib (dismissed):
Looks good! Thank you!

rstaib:
@tobias-tengler could you please fix the conflict then we could merge 👍

tobias-tengler:
@rstaib done! ✅

tobias-tengler:
Not sure why the Website pipeline also ran 17min this time. It seems to have been stuck on the onPostBuild, which this PR doesn't modify. I tested locally again and the link validation is definitely not at fault here. I would attribute it to degraded Pipeline performance of DevOps at the moment...

rstaib (approved):

rstaib:
@tobias-tengler Let's see if the perf issue is persistent and tackle it then in another issue.

tobias-tengler:
Happened again in the pipeline, while other PRs ran with a similar time. I'm not yet sure what's causing this, the link validation is already done, once the stalling hits. It might be related to the generation of the blog feeds. I'll look further into it tomorrow.

PR #3752, opened 2021-05-28
Autogenerate Strawberry Shake help texts

jorrit:
While running the Strawberry Shake dotnet tool I found that the help command listed options that were no longer valid.

The help texts were hardcoded in various `HelpTextGenerator`s and these generators were not updated when the command line options were modified. Furthermore, the DownloadCommand was referencing the help text generator from the InitCommand and two generators for non-existent commands were still present.

I've refactored the Program class and subcommands to use the built-in help text generation from McMaster. For some reason, this did not work when using the `AddSubcommand()` method to add subcommands, so I refactored it to use `Command()`.


jorrit:
Thanks!

PR #3797, opened 2021-06-04
Update website to gatsby v3

tobias-tengler:
Updates are now finally reflected in real-time again :rocket: 

### Changes

I upgraded all packages to their latest version and made the following replacements:
| old                               | new                           |
|-----------------------------------|-------------------------------|
| gatsby-plugin-ts                  | gatsby-plugin-graphql-codegen |
| gatsby-image                      | gatsby-plugin-image           |
| @darth-knoppix/gatsby-plugin-feed | gatsby-plugin-feed            |

The switch to `gatsby-plugin-feed` isn't necessary and I can revert it, if it's unwanted.

### Trying it out

1. Checkout the branch
2. Remove the existing `node_modules` directory and execute `yarn install` to install all necessary packages
3. Execute `yarn clean` to get rid of old (potentially problematic) cache entries

Continue as you normally would with `yarn start` or `yarn build`.


michaelstaib:
Uhhh Gatsby 3!

arif-hanif:
image

tobias-tengler:
@arif-hanif Can't do anything about the error, but it also doesn't cause any issues. The author information is supposed to be unknown during development. The actual git commands are only issued during a production build.

arif-hanif:
> @arif-hanif Can't do anything about the error, but it also doesn't cause any issues. The author information is supposed to be unknown during development. The actual git commands are only issued during a production build.
> I can add the dependency.

Looks good then.

arif-hanif (approved):
approved

website/gatsby-config.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
     },
   },
   plugins: [
-    `gatsby-plugin-ts`,
+    `gatsby-plugin-graphql-codegen`,
```
rstaib, line 38: This plugin produces the following warning: `warn [gatsby-plugin-graphql-codegen] Identifier 'getIterator' has already been declared (3564:9)`

website/gatsby-config.js
```diff
@@ -274,6 +274,7 @@ module.exports = {
                 };
               }),
             title: "ChilliCream Blog",
+            output: "/rss.xml",
```
rstaib, line 277: We have now dropped the support for `atom` and `json`. I guess because this plugin is outdated and no other plugin supports the mentioned formats, right?

rstaib (changes requested):
Pretty good! Thank you! I have discovered a few warning. Would be nice if you could get rid of them :-)

1. `warn Plugin gatsby-remark-reading-time is not compatible with your gatsby version 3.6.2 - It requires gatsby@^2.0.0`
2. `warn Query takes too long:
File path: /Users/raf/work/hotchocolate/website/src/templates/doc-page-template.tsx
URL path: /docs/strawberryshake/performance/persisted-queries
Context: {
    "originPath": "strawberryshake/performance/persisted-queries.md"
}
`
3. `warn [gatsby-plugin-graphql-codegen] Identifier 'getIterator' has already been declared (3564:9)`

Everything else looks decent 👍

website/gatsby-config.js
tobias-tengler: As I mentioned in the description, we don't have to switch to the other plugin. I just wanted to use less forks. If there are people using these feeds, we can definitely keep it around!

tobias-tengler:
The warn `Plugin gatsby-remark-reading-time is not compatible with your gatsby version 3.6.2`, doesn't cause any issues. In the migration documentation of Gatsby it also said that some packages will complain that they are not compatible, because their peer dependency was not upgraded, but most of them will continue to work. The plugin is just no longer maintained as it seems and therefor the peerDependency was not upgraded. I also couldn't fine an alternative.
The `Query takes too long` also sometimes happens with Gatsby v2.

website/gatsby-config.js
rstaib: I will resolve this one. But let's try targeting zero warning builds 😄

rstaib (approved):

PR #3839, opened 2021-06-16
Diagnose website CI performance issues

tobias-tengler:
# :warning: DO NOT MERGE :warning:


tobias-tengler:
@rstaib I couldn't pin down what was causing the increased build time after my "Catch dead links" PR, but here I've added it on top of the gatsby v3 update and the performance hit is gone: https://chillicream.visualstudio.com/HotChocolate/_build/results?buildId=26884&view=logs&j=1f1ae25b-b048-58e4-2d38-08d66b694a4a&t=ec2b6dca-c607-57dc-bdd5-71fcf5694d65
I think I won't look further into it and just work on getting the v3 update ready to merge.

PR #3844, opened 2021-06-19
WIP: Fix algolia search

tobias-tengler:
### Changes
- Moves the algolia indexing into the source code (I wasn't sure what was previously being indexed, so I just indexed docs/blog posts. I also used a throwaway algolia account to do so)
- Reduces the amount of data that is being returned from algolia --> less network strain
- Search results are now correctly shown as snippets

### TODOS
- Move the website CI definition into this repository
- Add the algolia admin API key as a GitHub secret and switch out all the keys currently checked in (remove admin key of dummy account from source code)
- The old search and this search does not index code snippets, but it would be really helpful, so I'll try and accomplish that
- Sometimes there are duplicate snippet results
- The old search provided anchors of headings to jump to (although they weren't always accurate), would be nice if this search would also contain this

Preview: https://chilly-algolia-search.netlify.app/

Note: Since this depends on the work I did in #3746 I also included it here, since it wasn't yet merged into develop and I couldn't do so myself. The diff is a bit noisy because of this, but this should reduces once main is merged into develop.


tobias-tengler:
@rstaib I'll need your help with these:
- Move the website CI definition into this repository
- Add the algolia admin API key as a GitHub secret and switch out all the keys currently checked in (remove admin key of dummy account from source code)

rstaib (changes requested):
Actually, you just need to modify the [config here](https://github.com/algolia/docsearch-configs/blob/master/configs/chillicream.json).

tobias-tengler:
@rstaib do I just create a PR to this repository and how do I test my changes!?

rstaib:
@tobias-tengler yes 👍

PR #3845, opened 2021-06-20
Fix for issue with spatial types

### LINKED ISSUE #3685: StrawberryShake Runtime errror deserializing GeoJSONPointType
Runtime error while deserializing the the 'coordinates' field of a GeoJSONPointType
HC and SS 11.3.0-preview.7

I'm trying to desterilize the 'coordinates' field of a point
```
fragment currentStatusFields on CurrentStatus {
    position {crs coordinates}
}
```
I'm getting this error:
```
The requested operation requires an element of type 'String', but the target element has type 'Array'., code:
   Exception: System.InvalidOperationException: The requested operation requires an element of type 'String', but the target element has type 'Array'
```

I assume that is because the actual JSON contains an array for the coordinates:
```
"position":{"__typename":"GeoJSONPointType","crs":4326,"coordinates":[-85.15869,41.12649,0]}
```

I've tried registering my own Serializer for the 'Position' scalar type (that's the type my schema reports for coordinates):
```
extend scalar Position
    @runtimeType(name: "global::NetTopologySuite.GeometriesGraph.Position")
    @serializationType(name: "global::System.Array")
```

However, in that case I get a build error:
```
Serialization format not supported
```

ChrisMH:
*PARTIAL FIX* for issue with using GeoJSON spatial types in GraphQL clients.

https://github.com/ChilliCream/hotchocolate/issues/3685

This is my first attempt at a contribution, and I wanted to make sure my approach was acceptable before continuing.  I have only changed GeoJSONPointType and GeoJSONLineStringType in the PR, but I'll continue with the rest if it looks OK.

I have had issues in both StrawberryShake and Apollo with the format of spatial data returned by HC.  The JSON looks fine, but I believe the schema is incorrect so the client can't properly generate code.

For all types, the 'coordinates' field uses a Position scalar.  However, Position should actually be an array of floating point values, not a scalar.

For example, the point type before:

```json
scalar Position

type GeoJSONPointType implements GeoJSONInterface {
  coordinates: Position
}
```

and the point type after:
```json
type GeoJSONPointType implements GeoJSONInterface {
  coordinates: [Float!]!
}
```

I have tested this change in a real scenario and it works as I would expect.

Closes #3685


PascalSenn:
@ChrisMH that is correct :) it should be an array of Float
the problem is that it's no possible to have N Dimensional Lists in graphql
The reason why we use Position is because some GeoJSON types would need [[[Float]]] which is invalid. 
We might even have to go a step further. as the MulitPolygon one is [[Position]] but should be [[[Position]]]
So we will even have to introduce another scalar at some point.

PR #3875, opened 2021-06-27
Fix UrlSerializer for absolute Urls

francisrath:
Current code always serializes Url as relative, but the proposed code serializes both relative and absolute Urls correctly.


michaelstaib (approved):

francisrath:
Any chance to get this merged for the next release?

PR #3902, opened 2021-06-30
Use lowercasing on the documentation for the filtering

fpaganetto:
When I was using HotChocolate on my own implementation I realized the only way to use the *or* and the *and* operators are with lowercase instead of the example that is on the documentation with uppercasing.

The displayed message is: **"The specified input object field `OR` does not exist."**

image

This is very confusing because taking the example literally will not work.

Summary of the changes

- Replace "AND" and "OR" with "and" and "or" using lowercasing


PascalSenn (approved):

tobias-tengler:
Thanks @fpaganetto for the contribution! :)

:star2: If you haven't already, please give the project a star - it helps tremendously :star2:

PR #4126, opened 2021-08-24
Added option to only allow persisted queries

### LINKED ISSUE #3943: Allow only persisted queries to be executed
While writing documentation for persisted queries (#3616) it came up that there is currently no good way of allowing only persisted queries to be executed.

There should be a new option in the server configuration users can set to refuse any "dynamic" queries and only execute requests which contain a query `id`:

```csharp
app.UseEndpoints(endpoints =>
{
    endpoints.MapGraphQL().WithOptions(new GraphQLServerOptions
    {
        AllowOnlyPersistedRequests = true
    });
});
```

`AllowOnlyPersistedRequests` is of course just an idea and nothing final.

PascalSenn:
Closes #3943


PascalSenn:
@michaelstaib  how do i have to fix the public api analyzer when the interface has changed? just remove the entries?

tobias-tengler:
@PascalSenn Yes. It's a bit tedious though. I've created a little python script a while back to go over the command output and remove all the RS0017 errors. I'll share it later today or try to improve the AddUnshipped to handle it automatically (I tried to do so before, but failed).

src/HotChocolate/Core/src/Execution/Pipeline/PersistedQueries/IPersistedQueryOptions.cs
```diff
@@ -0,0 +1,13 @@
+namespace HotChocolate.Execution.Pipeline
+{
+    /// <summary>
+    /// Represents common options for persisted queries
+    /// </summary>
+    public interface IPersistedQueryOptions
```
tobias-tengler, line 6: Any particular reason why you decided to introduce this as a separate interface? Conceptually the option to only allow persisted queries would live in the [`GraphQLServerOptions`](https://github.com/ChilliCream/hotchocolate/blob/c1eb36cf3542572372847c5e38f4bc349bc22bfa/src/HotChocolate/AspNetCore/src/AspNetCore/GraphQLServerOptions.cs) for me. Here you already configure what types of requests are allowed.

PascalSenn:
Follow up UseOnlyPersistedQueryPipeline

benmccallum (approved):
Awesome! 

Do you think this will fit into v12 @PascalSenn ? I guess worst case it'd be a v12.1 or something.

PascalSenn:
@benmccallum 
This should go into v12 i think

damikun:
@michaelstaib  V12.1 ? looks like forgotten pull.. :)

PR #4149, opened 2021-08-27
Adding new XmlDocumentationFileName configuration to ISchemaOptions a…

BlacKCaT27:

- Added new configuration option on `ISchemaOptions` and associated classes to allow for an optional file name of the xml documentation file to be provided. If none is provided, the existing behavior is maintained.


src/HotChocolate/Core/src/Types/Configuration/ReadOnlySchemaOptions.cs
```diff
@@ -63,6 +66,12 @@ public ReadOnlySchemaOptions(IReadOnlySchemaOptions options)
         /// </summary>
         public bool UseXmlDocumentation { get; }
 
+        /// <summary>
+        /// Defines the name of the XML documenatation file to be read.
+        /// Only used if <seealso cref="UseXmlDocumentation"/> is true.
+        /// </summary>
+        public string XmlDocumentationFileName { get; }
```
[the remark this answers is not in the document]
michaelstaib: I understand that, but if we put something in the public API it must cater to more than just one use case. I think in most case people have a filename that is based on the assembly. Having a delegate there would allow your use-case as well as all the others.

essentially in your use case you would just pass in `assembly => "MyStaticFileName.xml"`.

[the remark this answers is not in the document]
michaelstaib: Also respect nullability ... so the property should. be nullable.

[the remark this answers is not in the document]
BlacKCaT27: I'm merely an outside contributor so if you feel that's the approach the library should take I'll update the PR to implement that, but personally I feel requiring them to specify a delegate which takes an Assembly reference is more specialized than a string property. If they need/want to use the assembly name, they can pass `Assembly.GetExecutingAssembly().Name` or whatever else in their configuration code. Or is the concern that the code doing the initialization may not be the assembly that gets provided at runtime?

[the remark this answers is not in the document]
michaelstaib: Types in schemas can be spread to multiple assemblies. Many projects I have been working on have their types not only in the executing assembly. With the implementation you provided the name you provide would apply to all documentation files from all assemblies out of which you build the GraphQL server. If using the delegate you can dynamically generate that documentation file. In many cases it would be something like `assembly => assembly.GetName().Name + ".xml"`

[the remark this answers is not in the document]
michaelstaib: Looking at your comment it seems that you expect the naming convention to look only at a single assembly. It actually traverses all assemblies that are used to conceive a schema.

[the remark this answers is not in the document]
BlacKCaT27: Ah, thank you for that clarification. I went back and looked through the calling code and I see what you mean now about needing to be able to pull data from multiple assemblies. It didn't occur to me that the compiler didn't combine xml docs into a single file.

I'll update the PR to use a delegate so callers have access to the assembly in question at runtime.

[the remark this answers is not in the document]
michaelstaib: awesome

src/HotChocolate/Core/src/Types/Configuration/Contracts/IReadOnlySchemaOptions.cs
```diff
@@ -34,6 +36,12 @@ public interface IReadOnlySchemaOptions
         /// </summary>
         bool UseXmlDocumentation { get; }
 
+        /// <summary>
+        /// A delegate which defines the name of the XML documentatation file to be read.
```
michaelstaib, line 40:
```suggestion
        /// A delegate which resolves the name of the XML documentatation file to be read.
```

michaelstaib:
```suggestion
        /// A delegate which defines the name of the XML documentation file to be read.
```

src/HotChocolate/Core/src/Types/Configuration/Contracts/IReadOnlySchemaOptions.cs
```diff
@@ -34,6 +36,12 @@ public interface IReadOnlySchemaOptions
         /// </summary>
         bool UseXmlDocumentation { get; }
 
+        /// <summary>
+        /// A delegate which resolves the name of the XML documentatation file to be read.
```
michaelstaib, line 40:
```suggestion
        /// A delegate which resolves the name of the XML documentation file to be read.
```

src/HotChocolate/Core/src/Types/Configuration/Contracts/ISchemaOptions.cs
```diff
@@ -32,6 +36,12 @@ public interface ISchemaOptions : IReadOnlySchemaOptions
         /// </summary>
         new bool UseXmlDocumentation { get; set; }
 
+        /// <summary>
+        /// A delegate which defines the name of the XML documentatation file to be read.
+        /// Only used if <seealso cref="UseXmlDocumentation"/> is true.
+        /// </summary>
```
michaelstaib, line 42:
```suggestion
         /// <summary>
         /// A delegate which resolves the name of the XML documentation file to be read.
         /// Only used if <seealso cref="UseXmlDocumentation"/> is true.
         /// </summary>
```

src/HotChocolate/Core/src/Types/Configuration/ReadOnlySchemaOptions.cs
```diff
@@ -63,6 +67,12 @@ public ReadOnlySchemaOptions(IReadOnlySchemaOptions options)
         /// </summary>
         public bool UseXmlDocumentation { get; }
 
+        /// <summary>
+        /// A delegate which defines the name of the XML documentatation file to be read.
+        /// Only used if <seealso cref="UseXmlDocumentation"/> is true.
+        /// </summary>
```
michaelstaib, line 73:
```suggestion
         /// <summary>
         /// A delegate which resolves the name of the XML documentation file to be read.
         /// Only used if <seealso cref="UseXmlDocumentation"/> is true.
         /// </summary>
```

src/HotChocolate/Core/src/Types/Types/Descriptors/Conventions/DefaultNamingConventions.cs
```diff
@@ -21,16 +21,19 @@ public DefaultNamingConventions(IDocumentationProvider documentation)
...
 
-        public DefaultNamingConventions()
-            : this(true)
+        public DefaultNamingConventions() : this(true)
+        {
+        }
+        
+        public DefaultNamingConventions(bool useXmlDocumentation) : this(useXmlDocumentation, null)
         {
         }
```
michaelstaib, line 31: I reworked the constructors to keep binary compatibility for 11.

michaelstaib:
Hey everyone,

I have refined the PR a bit more and will merge it today. I have retargeted the change to 12 since it breaks interfaces. I amended some of the breaking changes but we cannot get this into 11. I will merge it into 12-rc.1 later today with a couple other fixes for 12.

michaelstaib:
One more thing this PR is missing is a test to prove the behavior.

src/HotChocolate/Core/test/Types.Tests.Documentation/HotChocolate.Types.Tests.Documentation.csproj
```diff
@@ -6,4 +6,10 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Update="$(MSBuildProjectDirectory)\Dummy.xml">
```
michaelstaib, line 10: I have added config with a custom name to our docs test assembly.

src/HotChocolate/Core/test/Types.Tests/Configuration/ReadOnlySchemaOptionsTests.cs
```diff
@@ -31,6 +31,33 @@ public void Copy_Options()
...
+                DefaultBindingBehavior = BindingBehavior.Explicit,
+                FieldMiddleware = FieldMiddlewareApplication.AllFields,
+                PreserveSyntaxNodes = true
+            };
+
+            // act
+            var copied = new ReadOnlySchemaOptions(options);
+
+            // assert
+            Assert.Same(
```
michaelstaib, line 56: We test that the options are correctly copied.

src/HotChocolate/Core/test/Types.Tests/Types/Descriptors/Conventions/XmlDocumentationProviderTests.cs
```diff
@@ -95,6 +95,22 @@ public void When_type_has_description_then_it_it_resolved()
             Assert.Equal("I am the most base class.", description);
         }
 
+        [Fact]
+        public void When_we_use_custom_documentation_files_they_are_correctly_loaded()
```
michaelstaib, line 99: This test will use the xml file with the custom name instead of the standard docs file.

src/HotChocolate/Core/test/Types.Tests/Types/Descriptors/Conventions/XmlDocumentationProviderTests.cs
```diff
@@ -95,6 +95,22 @@ public void When_type_has_description_then_it_it_resolved()
...
+            var documentationProvider = new XmlDocumentationProvider(
+                new XmlDocumentationFileResolver(_ => "Dummy.xml"),
+                new NoOpStringBuilderPool());
+
+            // act
+            var description = documentationProvider.GetDescription(
+                typeof(BaseBaseClass));
+
+            // assert
+            Assert.Equal("I am the most base class from dummy.", description);
```
michaelstaib, line 111: We check that we get the description from our modified config instead of the standard config. The standard config would state `I am the most base class.`. 

With this small integration test we have proved that our option really would work. 

The original test `When_type_has_description_then_it_it_resolved ` proves that the old behavior is still intact.

michaelstaib (approved):

PR #4255, opened 2021-09-21
add note on instrumentation

randyridge:
Add note about return types.


tobias-tengler (approved):
Thanks @randyridge  :)
I've also started to document instrumentation [here](https://github.com/ChilliCream/hotchocolate/blob/tte/instrumentation-docs/website/src/docs/hotchocolate/server/instrumentation.md) if you are interested!

PR #4282, opened 2021-09-28
Fix search on website

tobias-tengler:
I couldn't figure out how to solve the search issue on the website using the [docsearch config](https://github.com/algolia/docsearch-configs/blob/master/configs/chillicream.json), so I made it work using some code in the website.
This should at least be acceptable as a temporary fix.


arif-hanif:
@tobias-tengler thank you for looking at this, functioning much better, previous state was not useable at all

arif-hanif (approved):
approved

rstaib (approved):
Looks good to me. Thank you!

PR #4295, opened 2021-10-02
Website: Clean up support

arif-hanif:


rstaib (approved):
Good! One question regarding the deleted sales card component. Aren't we gonne need it for the BCP product page?

PR #4497, opened 2021-11-28
Document field middleware

tobias-tengler:


PascalSenn:
SWEET ❤️

This is on the list for ages, but no one had the time !! This is great! 



If you need images or inspiration you can get it from [Here]( https://github.com/ChilliCream/graphql-workshop/blob/master/docs/5-understanding-middleware.md ) :)

tobias-tengler:
Ah perfect 
image
is just what I needed! <3

website/src/docs/hotchocolate/execution-engine/field-middleware.md
```diff
@@ -0,0 +1,252 @@
...
+
+# Definition
+
+Field middleware can be defined either as a delegate or as a separate type. In both cases we gain access to a `FieldDelegate` (`next`) and the `IMiddlewareContext` (`context`).
+
+The `IMiddlewareContext` implements the `IResolverContext` interface so you can use all of the `IResolverContext` APIs in your middleware, similarly to how you would use them in your resolver code.
+
+By awaiting the `FieldDelegate` we are waiting on all other field middleware that might come after the current middleware as well as the actual field resolver, computing the result of the field.
+
+## Field middleware delegate
```
michaelstaib, line 19: Shall we also describe the request middleware for extending the execution as a whole? If you want I can write something up on that, I already have most of the diagrams for that.

tobias-tengler: I assume by request middleware you mean the pipeline stuff, i.e. `UseRequest`? I wanted to do a different document on that.

michaelstaib: Yes

tobias-tengler:
Not documenting local / scoped state at this point, as the APIs will likely change soon. I'm going to do the request pipeline and the resolver compiler next.

website/src/docs/hotchocolate/execution-engine/field-middleware.md
```diff
@@ -0,0 +1,291 @@
...
+
+```csharp
+public class UseMyMiddlewareAttribute : ObjectFieldDescriptorAttribute
+{
+    public override void OnConfigure(IDescriptorContext context,
+        IObjectFieldDescriptor descriptor, MemberInfo member)
+    {
+        descriptor.UseMyMiddleware();
+    }
+}
```
michaelstaib, line 226:
```suggestion
public class UseMyMiddlewareAttribute : ObjectFieldDescriptorAttribute
{
    public UseMyMiddlewareAttribute([CallerLineNumber] int order = 0)
    {
        Order = order;
    }

    public override void OnConfigure(IDescriptorContext context,
        IObjectFieldDescriptor descriptor, MemberInfo member)
    {
        descriptor.UseMyMiddleware();
    }
}

```

website/src/docs/hotchocolate/execution-engine/field-middleware.md
```diff
@@ -0,0 +1,291 @@
...
+    .Use((provider, next) => new MyMiddleware(next, "custom",
+        provider.GetRequiredService<FooBar>()));
+```
+
+# Usage as an attribute
+
+Up until now we have only worked with Code-first APIs to create the field middleware. What if you want to apply your middleware to a field resolver defined using the Annotation-based approach?
+
+You can create a new attribute inheriting from `ObjectFieldDescriptorAttribute` and call or create your middleware inside of the `OnConfigure` method.
+
```
michaelstaib, line 217:
```suggestion

> Note: Attribute order is not guaranteed in C# which is why we in the case of middleware attributes use the `CallerLineNumberAttribute` to inject at compile time the C# line number. The line number is used as an order. We do not recommend inheriting middleware attributes from a base field since this can lead to confusion about ordering. Look at the example below how we infer the order. When inheriting from middleware attributes always pass through the order argument. Further, indicate with the `Use` verb that your attribute is a middleware attribute.

```

PR #4561, opened 2021-12-14
feat: Add optional param to InitializeOnStartup

nelsonprsousa:

- Adds an optional param to control if GraphQL configuration should be added to the warmup background service.


PascalSenn:
I am not sure if I understand this correctly. 
This would just skip the initialisation on startup if it is true? 
What is the specific use case for this? 
Couldn't this be easily done with a extension method?

nelsonprsousa:
> I am not sure if I understand this correctly. This would just skip the initialisation on startup if it is true? What is the specific use case for this? Couldn't this be easily done with a extension method?

Yes, it will just skip the initialization on startup.

The main goal is for us to be able to easily configure this just like EF Core does:
image

This PR will allow us to do this:
image

nelsonprsousa:
@PascalSenn did you had the chance to take a look at this PR? Let me know if we can merge it or if we should close it. I am available to add/refactor/change something if it's necessary 👍

PR #4632, opened 2022-01-10
Update console.md

matthiasguentert:
Added note regarding build action


PascalSenn:
@matthiasguentert I cannot really validate that as i do not work with Visual Studio. @michaelstaib  do you still have a installation?

PR #4647, opened 2022-01-14
Update entity-framework.md

taspeotis:
As best I can tell the example did not implement what was indicated:

>  If you are using a custom attribute, you have to forward the line number using the `CallerLineNumberAttribute`.

So this passes the line number along.


PascalSenn (approved):

taspeotis:
Hi I must apologise, this is actually done immediately afterwards in the `Order = order` line.

PR #4661, opened 2022-01-18
Add note for xml docs + custom naming convention

benmccallum:


benmccallum:
There's currently no page for the naming conventions customization (since v10), but it's noted in the missing docs issue that Tobias made. 

When that page is added, it'd be good to cross-reference link to this note from that side too.

michaelstaib (approved):

michaelstaib (approved):

sgabler-solytic:
@benmccallum Thank you for documenting this edge case. It helped me a lot. But I think the documentation is not 100% up-to-date with v12.7.0 anymore

In this screenshot you can see that NoOpStringBuilderPool can't be instantiated:
image

If I remove the `?? new NoOpStringBuilderPool()`, the XML comments appear again. Only did a quick check though, so I don't know if there are any edge cases where this could become a problem

It seems that `NoOpStringBuilderPool` is private now, so that's why it can't be instantiated here.

Any idea how to best proceed here? Is my solution good enough, or should some sort of fallback still be implemented?

FYI @michaelstaib

PR #4752, opened 2022-02-11
Update Gatsby and fix scroll issue

tobias-tengler:
A user reported the inability to scroll using arrow keys, with this change the website can be scrolled again, albeit the scrollbar is now partially hidden behind the header.


tobias-tengler:
@rstaib Is it intentional that the .yarn directory with its cache was checked in?

website/src/components/misc/global-style.tsx
```diff
@@ -12,7 +12,7 @@ export const GlobalStyle = createGlobalStyle`
 
     --border-radius: 4px;
     --font-size: .833rem;
-  }
+  }
```
rstaib, line 15: empty space

website/src/components/misc/global-style.tsx
```diff
@@ -35,12 +35,10 @@ export const GlobalStyle = createGlobalStyle`
 
     > div:first-child {
       display: block;
-      height: 100vh;
```
rstaib, line 38: removing height leads to another issue, where navigation does not take anymore the entire height of the content area.

website/src/components/misc/global-style.tsx
```diff
@@ -35,12 +35,10 @@ export const GlobalStyle = createGlobalStyle`
...
 
       > div {
         display: flex;
         flex-direction: column;
-        height: 100vh;
```
rstaib, line 43: same here

rstaib (changes requested):
introduces another issue in regard to navigation height

PR #4877, opened 2022-03-19
Adds expression support for filtering

PascalSenn:
Adds expression support for filtering


michaelstaib (approved):

Jonatthu:
Why not to produce instead an array of the enum on the graphql side? 

Like:

'''
myQuery(where { type: [Flag1, Flag2] }) {...}
'''

PR #4882, opened 2022-03-20
WIP: Add failing tests for extending Code-First types

tobias-tengler:
While playing around with Type Extension I've noticed that the following two cases do not work:

Extending a Code-first type from the Annotation-based approach:
```csharp
public class CodeFirstType : ObjectType
{
    {
        descriptor.Name("Query");

        descriptor.Field("field1").Resolve("");
    }
}

[ExtendObjectType(typeof(CodeFirstType))]
public class ExtendCodeFirstType
{
    public string Field2 => "";
}
```

Extending a Code-first type using `ExtendsType<T>()`:
```csharp
public class CodeFirstType : ObjectType
{
    {
        descriptor.Name("Query");

        descriptor.Field("field1").Resolve("");
    }
}

public class CodeFirstTypeExtension : ObjectTypeExtension
{
    {
        descriptor.ExtendsType<CodeFirstType>();

        descriptor.Field("field2").Resolve("");
    }
}
```

I wasn't sure whether these are intentional limitations or whether they are bugs and how to best go about solving them. In the end I just added these tests to showcase the issue.

### DIFF
--- a/src/HotChocolate/Core/src/Types/Configuration/TypeInitializer.cs
+++ b/src/HotChocolate/Core/src/Types/Configuration/TypeInitializer.cs
@@ -231,6 +231,8 @@ internal class TypeInitializer
             _globalComps,
             _isOfType);
 
+        // todo: this fails, when using the ExtendsType<T> API
+        //       without specifying a name for the object type extension.
         registeredType.Type.CompleteName(registeredType);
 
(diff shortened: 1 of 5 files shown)


tobias-tengler:
Cool. I don't think this is a priority, as nobody else has reported or run into this. Closing for now as I don't plan on pursuing this further.

PR #4970, opened 2022-04-15
WIP: Adds Projection Providers for EF Core 

PascalSenn:
Adds a new extensions for projections to the request executor builder:
```
.AddEntityFrameworkProjections()
```

This extension methods registers a projection provider specifically made for EntityFramework.  In case the projected queryable is a `EntityQueryable<TEntityType>` the nullchecks in the projections are disabled. This way ef core can also detect projections that are first mapped with a `Select(x => ...)`


src/HotChocolate/Data/test/Data.EntityFramework.Tests/IntegrationTests.cs
```diff
@@ -462,4 +466,76 @@ public async Task ExecuteAsync_Should_ReturnNull_When_FirstOrDefaultZero_AsyncEn
...
+                                {
+                                    Id = book.Author!.Id,
+                                    Name = book.Author.Name
+                                }
+                            })
+                    )
+                    .UseFirstOrDefault()
+                    .UseProjection()
+                    .UseFiltering()
+                    .UseSqlLogging()
```
kresimirlesic, line 499: @PascalSenn I pulled this branch and did some debugging - this change is exactly what will resolve the EF issue.

I think your test is falsely failing - once you move this "UseSqlLogging" call before the "UseProjection" call you should get the expected sql and expression.

kresimirlesic:
@PascalSenn there is only one caveat with this approach. 
The inner join is working correctly and as expected, but the left join is going to throw an exception if the entity does not exist in the database.  This would require the queryable in the resolvers to be manually changed a bit. Personally, I can live with this change for now as a quick fix, because I think the inner joins and overselecting are a lot bigger issue and more frequent than the left joins. 

I would really like to see this change go inside, but more as a hotfix than a long term solution, because I think the `Key()` API (or something similar) as @AradAlvand suggested in [this issue](https://github.com/ChilliCream/hotchocolate/issues/2373) is probably the long term and 100% working solution. I am not sure if you and the team are ready to commit to the `Key` API or how long it will take you to develop something like that, so while we wait for that I think this would be a somewhat OK change.


For example what will we have to do for left joins....

From
```
new BookDto
{ 
        Title = book.Title, 
        Author = book.Author.Id != null 
              ?  new AuthorDto
                  { 
                          Id = book.Author.Id, 
                          Name = book.Author.Name
                  }
              : null <--------------
}
```

to

```
new BookDto
{ 
        Title = book.Title, 
        Author = book.Author.Id != null 
              ?  new AuthorDto
                  { 
                          Id = book.Author.Id, 
                          Name = book.Author.Name
                  }
              : new AuthorDto() <-----------
}
```

kresimirlesic (approved):

PascalSenn:
@kresimirlesic  why does this throw an exception? can you provide a failing unit test?

kresimirlesic:
> @kresimirlesic why does this throw an exception? can you provide a failing unit test?

@PascalSenn 
if you copy paste your new test and do this projection instead, you will get the same exception as if there are no records in the left join.

```
.Select(book =>
    new BookDto
    {
        Title = book.Title,
        Author = null <--- THIS
    })
```

PascalSenn:
@kresimirlesic  I've added a failing tests
image

This solution will not work too btw:
```csharp
new BookDto
{ 
        Title = book.Title, 
        Author = book.Author.Id != null 
              ?  new AuthorDto
                  { 
                          Id = book.Author.Id, 
                          Name = book.Author.Name
                  }
              : null <--------------
}
``` 
as `book.Author.Id != null` will never be false

AradAlvand:
@PascalSenn 

Let me try to recap what's happening here.
When you said:

> as book.Author.Id != null will never be false

It will never be false if every book necessarily has (i.e. belongs to) an author, which is of course the case in this domain. But in cases where the relationship isn't like that the null check would be necessary.

An example of that would be a `Comment` => `User` relationship where the `UserId` foreign key in `Comment` is optional, since comments could be posted by anonymous users, let's say, so a comment can potentially NOT belong to a user.

@kresimirlesic is right in saying that removing the `NotNullAndAlso` (aka the null check) would solve the overfetching problem in the former case, but in the latter, we definitely do need a null check.

Let's take the `Comment`/`User` case as an example where the user is optional and so we need to have a null check. Now, of course without an intermediary projection, HC could do a simple null check on `User`, and EF Core will generate the expected SQL, but when we do have imterdiary projections (like DTOs), if HC does this:

```csharp
c => new CommentDto {
    User = c.User == null ? default : new UserDto { ... },
};
```
The null check `c.User == null` results in EF Core fetching the whole `User`. But HC instead does the null check on the key property, EF Core will generate the right SQL, and no overfetching will occur:
```csharp
c => new CommentDto {
    User = c.User?.Id == null ? default : new UserDto { ... },
};
```

I hope this clears things up. Let me know if anything's still confusing.

PascalSenn:
> Well, it will never be false (just as book.Author != null will never be false) if every book necessarily has (i.e. belongs to) an author, which is of course the case in this domain. But in cases where the relationship isn't like that the null check would be necessary.

This one is on me, i didnt push the latest changes. There is an example where the author is nullable

> The null check c.User == null results in EF Core fetching the whole User. But if HC instead does the null check on the key property, EF Core will generate the right SQL, and no overfetching will occur:

```csharp
new CommentDto {
    User = c.User?.Id == null ? default : new UserDto { ... },
}
```

This is not a valid expression. Null Conditionals are not allowed in expressions.

AradAlvand:
> This is not a valid expression. Null Conditionals are not allowed in expressions.

You're right. I'd forgotten that. But `c.User.Id == null` works too. The point stands.
Not in memory of course, only for EF Core, because you would obviously get a null exception if you try to run this in memory and `c.User` is null.

PascalSenn:
hmm. interesting. Ugly hack, but the SQL seems to be correct. It still selects "Id is NULL" but the projection is applied.. 

Theoretically we can get the key form the DB context. But there are several problems with this.. We cannot access the Model from the EntityQueryable as it is wrapped in the Compiler as a private field. 
Getting the Db Context from the context data is also not really an option as we support multiple db context in a resolver..
Did anyone already reach out to the EF team and ask why they do it this way? The would have all the necessary data available to make this decision. We would have to hack our way around it

PascalSenn:
This PR is not the right approach. We either find a good way to provide the dbcontext or it is fixed upstream. the latter is preferable as the EF query compiler has all necessary infromation available

PR #5033, opened 2022-05-05
Mof/elasticsearch string contains filter operation

frzmohammadali:

Added filter operations:
- string `contains`
- string `ncontains`

Note:  The elasticsearch `WildcardQuery` is used for the mentioned filter operations as of my current knowledge about elastic queries. Is there any better choice that you could think of?


src/HotChocolate/ElasticSearch/src/Data.ElasticSearch.Driver/Driver/ContainsOperation.cs
```diff
@@ -0,0 +1,20 @@
+namespace HotChocolate.Data.ElasticSearch;
+
+public class ContainsOperation : ILeafSearchOperation
+{
+    public ContainsOperation(
```
PascalSenn, line 5: Can we rename this to WildcardOperation to align with elastic?

src/HotChocolate/ElasticSearch/src/Data.ElasticSearch.Driver/Filters/Handlers/ElasticSearchStringContainsOperationHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        object? parsedValue)
+    {
+        if (parsedValue is not string val)
+        {
+            throw ThrowHelper.Filtering_WrongValueProvided(field);
+        }
+
+        IElasticFilterMetadata metadata = field.GetElasticMetadata();
+
+        return new ContainsOperation(
```
PascalSenn, line 49: Can we do 
```
Value = $"*{operation.Value}*"
``` 
here isntead of the rewriter?
If we do it in the rewriter we have to do it twice, once of elastic and once for opernsearch. lets just do it here once. 
Also, We have to escape the value. So when someone provides `where: {name: {contains: "foo*bar"}}` then we have to rewrite the value to "foo\*bar".
Possible characters to escape `+ - = && || > < ! ( ) { } [ ] ^ " ~ * ? : \ /` 
we need to check what we need to escape for the `Wildcard` operation :)

PascalSenn (changes requested):
Look good aswell @frzmohammadali ! 
Have a couple of questions :) 

and sorry that it took so long to review. try to to it quicker in the future :)

src/HotChocolate/ElasticSearch/src/Data.ElasticSearch.Driver/Driver/ContainsOperation.cs
frzmohammadali: Sure thing! 
At the time of coding, it got me thinking for a while, whether I should take HotChocolate naming conventions as the first priority or elastic queries and I ended up choosing contains. ;)
I'll rename this and consider paying more attention to elastic naming conventions in the future.

src/HotChocolate/ElasticSearch/src/Data.ElasticSearch.Driver/Filters/Handlers/ElasticSearchStringContainsOperationHandler.cs
frzmohammadali: Good point! Thanks for noticing ;)

I need a bit more explanations though. First of all, according to [elastic query documents](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html), the query supports 2 wildcard operations: `*` and `?`. Therefore I think we only need to escape these two, right?

Secondly, if someone provides `where: {name: {contains: "foo*bar"}}`, to my understanding, we need to escape the `*` in the middle of the string so that it gets treated as the actual asterisk character and still add the two `*` in the beginning and the end of the string. To what extend is this algorithm true?

PascalSenn: yes that seems right. But keep in mind, we also have to escape the escpae character `\`

src/HotChocolate/ElasticSearch/src/Data.ElasticSearch.Driver/Driver/ContainsOperation.cs
PascalSenn: The idea is that the "Operations" in the core package are as close as possible to the elastic search representation. It should really just be a 1:1 mapping.

PR #5048, opened 2022-05-11
Always use NtsGeometryServices instance for geometry creation

wassim-k:
[NetTopologySuite Getting Started](https://github.com/NetTopologySuite/NetTopologySuite/wiki/GettingStarted) recommends modifying `NtsGeometryServices.Instance` with the your own desired defaults.

Unfortunately, HotChocolate currently ignores those defaults by calling constructors of NTS Geometry classes directly instead of using `GeometryFactory`

This PR is a demonstration of the required change in order to respect the defaults set by `NtsGeometryServices.Instance` and I guess a way to start the discussion.

Thoughts?


PascalSenn (approved):

PascalSenn (approved):

PascalSenn:
@wassim-k  Sorry that it took so long to review this one! 
Great work, very clean!

Just fixed the CRS Resolver as now the default CRS is -1. I guess we anyway have an issue in this resolver we have to address at some point. But for now, the behaviour stays the same at least ;) 
I also optimized the Postgres integration tests.

Great work, thanks for the contribution. LGTM

PR #5122, opened 2022-06-04
Fixed total count breaking change in connection;

PascalSenn:

- Detail 1
- Detail 2


src/HotChocolate/Core/src/Types.CursorPagination/Connection.cs
```diff
@@ -54,7 +54,7 @@ public class Connection : IPage
         ConnectionPageInfo info,
         int totalCount = 0)
     {
-        _getTotalCount = _ => new(totalCount);
+        _getTotalCount = _ => new(info.TotalCount ?? totalCount);
```
PascalSenn, line 57: We actually never used total count before. I know use either total count of the connection page info or the one provided directly through the connection.

michaelstaib (approved):

PR #5295, opened 2022-08-10
Extended documnetation on Subscriptions for v12

cts-tradeit:
#5288
#5281

Besides explaining in detail how subscription work and what makes SubscribeAndResolve differ from Subscribe added new section on using Subscriptions without HC's pubsub.

### DIFF
--- a/website/src/docs/hotchocolate/defining-a-schema/subscriptions.md
+++ b/website/src/docs/hotchocolate/defining-a-schema/subscriptions.md
@@ -256,6 +256,8 @@ public async Book PublishBook(Book book, [Service] ITopicEventSender sender)
 
 If more complex topics are required, we can use the `ITopicEventReceiver`.
 
+> ⚠️ Note: Pay attention to the use of `SubscribeAndResolve` instead of `Subscribe`. As can be seen in the code-first example above, all subscription fields are in fact made of two methods (resolver and subscriber). When using `Subscribe`, the attributed method represents resolver method (with subscriber method generated at runtime) whereas `SubscribeAndResolve` denotes subscriber method (with resolver method generated at runtime).
+
 ```csharp
 public class Subscription
 {
@@ -277,9 +279,10 @@ public async Book PublishBook(Book book, [Service] ITopicEventSender sender)
 }
 ```
 
-If we do not want to mix the subscription logic with our resolver, we can also use the `With` argument on the `Subscribe` attribute to specify a seperate method that handles the event subscription.
+If we do not want to mix the subscription logic with our resolver, we can define both resolver and subscriber methods separately. Resolver method can point to the subscriber method using `With` argument on the `Subscribe`.
 
 ```csharp
+
 public class Subscription
 {
     public ValueTask<ISourceStream<Book>> SubscribeToBooks(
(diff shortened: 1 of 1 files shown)


website/src/docs/hotchocolate/defining-a-schema/subscriptions.md
```diff
@@ -256,6 +256,8 @@ public async Book PublishBook(Book book, [Service] ITopicEventSender sender)
 
 If more complex topics are required, we can use the `ITopicEventReceiver`.
 
+> ⚠️ Note: Pay attention to the use of `SubscribeAndResolve` instead of `Subscribe`. As can be seen in the code-first example above, all subscription fields are in fact made of two methods (resolver and subscriber). When using `Subscribe`, the attributed method represents resolver method (with subscriber method generated at runtime) whereas `SubscribeAndResolve` denotes subscriber method (with resolver method generated at runtime).
```
tobias-tengler, line 259: For the SubscribeAndResolve part I think it would be more understandable, if you write something like "each event message is directly returned as the subscription resolver result".

tobias-tengler (changes requested):
Thank you for your contribution :)
I've left some feedback, once addressed we can get this in :)

website/src/docs/hotchocolate/defining-a-schema/subscriptions.md
```diff
@@ -290,4 +293,54 @@ public class Subscription
...
+        [Service] EventHub hub)
+    {
+        IAsyncEnumerable<BookCreated> messageStream = hub.ListenAsync<BookCreated>();
+        await foreach(BookCreatedEventArgs message in stream) {
+            Book book = message.Book;
+            if (@book.Author == author)
+                yield return book;
+        }
+
+        return receiver.SubscribeAsync<string, Book>(topic);
```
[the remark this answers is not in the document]
cts-tradeit: fixed

website/src/docs/hotchocolate/defining-a-schema/subscriptions.md
```diff
@@ -290,4 +293,54 @@ public class Subscription
...
+
+    [Subscribe(With = nameof(SubscribeToBooks))]
+    public Book BookAdded(string author, [EventMessage] Book book)
+        => book;
+}
+```
+
+> ⚠️ Note: Arguments and return type of nodes in the schema is inferred from the resolver method. As in example with attribute `Topic`, all arguments must be be defined on resolver method even if accessed only by subscriber method.
+
+> ⚠️ Note: **Name and type of the parameter must match otherwise** subscriber method won't be called and **subscription will fail silently**. Some types like `int` and `long` may work with each other, but if and only if the value is in range.
```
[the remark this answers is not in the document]
cts-tradeit: fixed

website/src/docs/hotchocolate/defining-a-schema/subscriptions.md
cts-tradeit: What do you mean? Each iteration of the IAsyncEnumerable is followed by incovation of the resolver, in this case the "shallow" resolver looks like ctx => ctx.GetMsg<Foo>() so it just directly returns the message.

It could mention the shape of the compiled delegate, but I do not think it wise to skip the part where this method's body is in fact the subscriber method (as seen in SubscribeAndResolveAttribute's code). It also explains why UseProjectionAttribute won't work here (as it is not a resolver method).

cts-tradeit: @tobias-tengler

tobias-tengler (approved):

cts-tradeit:
@michaelstaib what am I supossed to do with the cla bot? or more precisely why is it blocking this request exclusively? i did not see it in others

cts-tradeit:
or @PascalSenn or @rstaib can you provide info regarding CLA?

michaelstaib:
I will send you the cla to sign on github... sorry for the delay.

PascalSenn:
@cts-tradeit  we found a easier solution ;) just click on the link provided by @CLAassistant

michaelstaib:
@cts-tradeit  we have now everything setup for the CLA, click on the link in the CLA bot message and sign the agreement.

Deathrage:
I signed CLA for both accounts. Can someone else do the second approval?

**Some feeback:** This agreement does not take conflict of laws into account, which might pose serious issue when all parties are not from a single country (I'm not Siwss, nor are many others). This raises question which law is applicable and which forum (understand court) has jurisdiction. It is wise to specify this in the contracts as applicable law and forum can be chosen in generall by the parties. This does not matter in the EU (and EEA in general) that much as regulations such as Rome I and Brussels I bis set unified coflict of law rules but even though Switzerland generally adopts EU regulation, these ones were not adopted. Without explicitng chocie, the applicable law and forum are determined by bilateral treaties and if no treaty exist then by both state's law concurently. Explicitly stating applicable law and forum might save you some headache one day.

michaelstaib:
ChilliCream Inc is registered in Delaware, US. So, US contract law applies.

image

We defined the company as project entity, so the contract is signed with the project entity. We took over the contract from the Linux Foundation.

michaelstaib:
Thank you for the contribution.

michaelstaib:
@Deathrage I fixed it ... We are now following the SAP CLA which explicitly states the court of law and stuff. The CLA Bot will ask people to resign over the next couple of days.

In the end we try to ensure that everything stays MIT and we do not get people that contribute and then file copyright claims against our users :)

Its not my area of expertise, so thank you in any case for pointing this out.

this is the new one:
https://gist.github.com/michaelstaib/371d461371f0ffc78dd6039e08b180fc

Deathrage:
@michaelstaib can you cherry-pick the this merge to main? (or I can create another PR)

Deathrage:
When this change will propagate to the live website?

PR #5474, opened 2022-10-09
Improve documentation navigation

tobias-tengler:
TL;DR:
- Improves the UX of navigating between documentation versions on the website
- Makes it easier and less error-prone to add new documentation versions in the future

## Changes

- Fixes broken images / components caused by incorrect relative paths
  - All versions of the product documentation are now in sub-directories, so this will no longer be an issue when creating new documentation versions in the future.
- Introduce new heading for preview documentation and update heading of outdated documentation:
  - [Example of preview docs](https://chilly-docs-navigation.netlify.app/docs/hotchocolate/v13/get-started)
  - [Example of outdated docs](https://chilly-docs-navigation.netlify.app/docs/hotchocolate/v11/get-started)
- Clicking the "latest stable version" link in the above warnings or using the version switcher now tries to open the document you're currently viewing for the new version instead of the starting page of the new version.
- Update the 404 page for documentation pages [Example](https://chilly-docs-navigation.netlify.app/docs/hotchocolate/v13/whatever)
- Links of the stable version now also include a version identifier, so links that are posted around are sure to link to the correct content.
- Added [documentation](https://github.com/ChilliCream/hotchocolate/blob/tte/improve-docs/website/README.md) on how to work with the documentation 😅
- Removed the Green Donut documentation, since it's really outdated and now covered by the Hot Chocolate DataLoader documentation.


rstaib (approved):
LGTM

PR #5493, opened 2022-10-20
Slice source text in SyntaxException

PascalSenn:
We were allocating a string with the whole source text in case of a syntax exception.

This is a problem when we parse SDLs

This PR allocates at max a string of 512 characters and also provides the offset to the original position


src/HotChocolate/Language/src/Language.Utf8/SyntaxException.cs
```diff
@@ -28,4 +30,37 @@ internal SyntaxException(Utf8GraphQLReader reader, string message, params object
     public int Column { get; }
 
     public string SourceText { get; }
+
+    public int SourceTextOffset { get; set; }
```
PascalSenn, line 34:
```suggestion
    public int SourceTextOffset { get; }
```

PascalSenn:
@michaelstaib  done

michaelstaib (approved):

PR #5530, opened 2022-11-03
Redirect v10 documentation links

### LINKED ISSUE #5529: Switching between documentation versions results in 404

### Describe the bug

Lots of google search results point to older documentation version, e.g. v10. When you try to switch to newer version, e.g. v12, the website just throws you to 404, which is absolutely frustrating. This happens because the website just replaces "v10" folder with "v12", and since there is no such page, you get 404. For example:

https://chillicream.com/docs/hotchocolate/v10/data-fetching -> switch to v12 -> redirect to https://chillicream.com/docs/hotchocolate/v12/data-fetching -> 404

However there is a page at https://chillicream.com/docs/hotchocolate/v12/fetching-data

Same bug about yellow message about deprecated version:

```
This is documentation for v10, which is no longer actively maintained.
For up-to-date documentation, see the latest stable version.
```

"latest stable version" is a link which also points to 404.

You should add additional URL mapping between different versions to same pages (or most close ones if there is no exactly same page in the newer version).

### Steps to reproduce

1. Navigate to https://chillicream.com/docs/hotchocolate/v10/data-fetching
2. Change version to v12 OR click "latest stable version" link in the yellow message block
3. 404

### Relevant log output

_No response_

### Additional Context?

_No response_

### Product

Hot Chocolate

### Version

12

tobias-tengler:
- Redirects all Hot Chocolate v10 links (that are covered) to their equivalent document in the latest version
- Adds the version of the product to the HTML title (this should hopefully help with SEO)
image

Closes #5529


rstaib (commented):
Hey @tobias-tengler,

sorry for letting you wait on this one and thank you for the work.

There is one little thing regarding how the redirects are created.

website/gatsby-node.js
```diff
@@ -329,6 +337,164 @@ function createDocPages(createPage, data, products) {
   });
 }
 
+function createHotChocolateRedirects(product, createRedirect) {
```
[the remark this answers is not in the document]
tobias-tengler: The automatic redirects are already being done. These are the ones that have completely changed locations so they need to be explicitly redirected somewhere else. Don't think we can make this any more dynamic.

website/src/components/doc-page/doc-page.tsx
```diff
@@ -204,23 +205,37 @@ function useProductInformation(
...
   return {
-    name: selectedName,
+    path: selectedPath,
+    name: getProductNameFromPath(selectedPath),
     version: selectedVersion,
     stableVersion,
   };
 }
 
+function getProductNameFromPath(path: string): string | null {
```
rstaib, line 226: Why not taking the `title` instead of the `path`?

tobias-tengler: What do you mean? This method is used to derive the name from the path segment and to later set the window title. So injecting the window title here wouldn't work, since we don't have it yet.

website/gatsby-node.js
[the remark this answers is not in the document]
rstaib: okay got it 👍🏻

website/src/components/doc-page/doc-page.tsx
rstaib: My concern here is, that we would have to maintain the product names in two different places, whereas we could achieve to manage these names in one place instead by reading these from the JSON file.

rstaib (approved):
Nice!

PR #5590, opened 2022-12-08
fix typings for React

artola:
Use resolution to align used version of `@types/react` and `@types/react-dom`.

Note: because transitive dependencies require that packages as `peerDependencies` using `*` for version, it leads to resolve to a different version (v18) than the actual required by us (v17).


rstaib (approved):
LGTM

PR #5598, opened 2022-12-11
Update mutations.md

sarvasana:
Some of the types mentioned in the documentation did not exist in the examples.
Replaced them with types present on the same page.

Also the factory was returning specialized types instead of a common base type.


tobias-tengler (approved):
Thank you :)

PR #5617, opened 2022-12-17
enhance seo

artola:
- add "robots.txt"
- add structured data snippet
- avoid prefetch failure for non-existing pages


rstaib (approved):
LGTM

PR #5618, opened 2022-12-18
enhance seo

artola:
- reduce index page size from 403K to 120K
- use sprite for big svg (artwork, companies, logo)
- refactor imports using alias


website/src/components/layout/main.tsx
```diff
@@ -1,11 +1,12 @@
...
 import styled from "styled-components";
-import { hasScrolled } from "../../state/common";
-import { PageTop } from "../misc/page-top";
+
+import { PageTop } from "@/components/misc/page-top";
+import { hasScrolled } from "@/state/common";
 import { Footer } from "./footer";
 
-export const Main: FC<PropsWithChildren> = ({ children }) => {
+export const Main: FC<PropsWithChildren<unknown>> = ({ children }) => {
```
rstaib, line 9: why `unknown`?

website/src/components/mdx/tabs/tab-groups.tsx
```diff
@@ -26,7 +26,9 @@ const TabGroupContext = createContext<TabGroupContextShape>({
 
 const getLocalStorageKey = (groupId: string) => `tab-${groupId}`;
 
-export const TabGroupProvider: FC<PropsWithChildren> = ({ children }) => {
+export const TabGroupProvider: FC<PropsWithChildren<unknown>> = ({
```
rstaib, line 29: same here

rstaib (approved):
LGTM

website/src/components/layout/main.tsx
artola: because in the version of typings used (v17) we are forced to pass `<P>` ... while in the future version v18, it defaults to `unknown` if you are not passing it.

website/src/components/mdx/tabs/tab-groups.tsx
artola: same than above, there were 4 or 5 places with the same ... once we update to v18 types, we will remove them.

website/src/components/layout/main.tsx
rstaib: Yeah, PropsWithChildren isn't even the best way to go and if I remember correct should be replaced.

PR #5621, opened 2022-12-19
enhance seo

artola:
- reduce index page size from 122kB to 117kB
- use sprite for brands svg
- standardise sprite (artwork, brands, companies, logo)
- use the same order for brand links in footer and header


rstaib (approved):
LGTM

PR #5625, opened 2022-12-20
enhance website

artola:
- add support to open mdx external links in a new page
- add sprite attach point
- normalise title case (NYT)
- improve header for small mobile
- change element for headings in homepage


website/plugins/gatsby-plugin-svg-sprite/gatsby-node.js
```diff
@@ -1,42 +1,31 @@
...
-          ) {
-            return {
-              ...item,
-              test: /\.(ico|jpg|jpeg|png|gif|webp)(\?.*)?$/,
-            };
-          }
-
-          return { ...item };
-        }),
+  config.module.rules = [
```
rstaib, line 11: are we sure that this has no side-effects?

rstaib (approved):
LGTM (Just one things)

website/plugins/gatsby-plugin-svg-sprite/gatsby-node.js
artola: > are we sure that this has no side-effects?

this is the same done in all these kind of plugins, for example see: [gatsby-plugin-react-svg](https://www.gatsbyjs.com/plugins/gatsby-plugin-react-svg/#gatsby-plugin-react-svg-npm-version)

https://github.com/jacobmischka/gatsby-plugin-react-svg/blob/master/gatsby-node.js

PR #5712, opened 2023-01-23
Prevent canceled @stream tasks from breaking subsequent @stream query

atykhyy:
This fix prevents canceled `@stream` tasks from breaking subsequent `@stream` queries.

When a query containing deferred `@stream` is canceled (e.g. due to http connection closing), `DeferResultStream.GetAsyncEnumerator()` disposes the work state owner, invoking `DeferredWorkState.Reset()`, before the tasks waiting for items from async enumerators being `@stream`'ed are canceled. When these tasks are canceled, they add `IQueryResults` with `Error`s set to `OperationCanceledException` to the freshly reset `DeferredWorkState`. The next query which rents this `DeferredWorkState` picks up these stale results and quits in `DeferredWorkState.TryDequeueResultsAsync()` after sending its initial result, so that the client invoking the query receives an empty result with `hasNext:false` and no `@stream` items.


michaelstaib (approved):

atykhyy:
Warning: this doesn't fix the problem completely but it does reduce the window of the underlying race condition very substantially. Instead of every other subscription being truncated as described above, I now observe truncation rarely and sporadically.

PR #5834, opened 2023-02-13
Fix wrong code in operation complexity documentation

sgabler:

- Fix code in documentation, as it didn't compile for me

image

I am not sure if this change should also be applied to other versions of the documentation


tobias-tengler (approved):
Thank you :)

michaelstaib (approved):

PR #5989, opened 2023-03-25
Optimize dataloader fetching for lists

PascalSenn:
Dataloader currently have a issue with batching.

The first call to "LoadAsync" will initiate a `Schedule` on the batch dispatcher. In a specfic combination of cases, a fetch of a list of keys will lead to two fetches instead of one. 

1. The flag "dispatchOnSchedule" in the batch dispatcher is true (e.g. Mutation Resolver)
1. You fetch more than one 1. key

What happens is the following. While we are iterating through all the requested keys, the first key will call schedule on the batch dispatcher. When we are calling `LoadAsync` from a mutation, `Schedule` will immediately dispatch. We continue synchronously and enter reenter the lock of the dataloader that we arleady own. We immediatly fetch with just one key instead of all the keys.


michaelstaib (approved):

michaelstaib (changes requested):
The changes lead to dead locks.

PR #6029, opened 2023-04-05
Fix issue `JSON` scalar crashes on JsonElement inputs #6023

### LINKED ISSUE #6023: `JSON` scalar crashes on JsonElement inputs

### Product

Hot Chocolate

### Describe the bug

When an input is of type `JSON`, HotChocolate crashes for numeric inputs (such as `10`), with the following exception

```
System.Text.Json.JsonReaderException: '0x00' is an invalid end of a number. Expected a delimiter. LineNumber: 0 | BytePositionInLine: 6.
```

### Steps to reproduce

1. Create an input with a property of type `JsonElement` in HC13
```C#
[QueryType]
public static class JsonBasedQuery
{
    public static string GetDataWithFilter(FilterType input)
    {
        return "hello";
    }

    public class FilterType
    {
        public string? Name { get; set; }
        public System.Text.Json.JsonElement Value { get; set; }
    }
}
```

2. Invoke it with a number literal
```graphql
query {
  dataWithFilter(input: {
    name: "hello"
    value: 10
  })
}
```

### Relevant log output

      GraphQL service exception for 'POST /graphql/UnknownOperation - 4a22f1afa6b4662d460ab8929d3c10f6': '0x00' is an invalid end of a number. Expected a delimiter. LineNumber: 0 | BytePositionInLine: 2.
      System.Text.Json.JsonReaderException: '0x00' is an invalid end of a number. Expected a delimiter. LineNumber: 0 | BytePositionInLine: 2.
         at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)        
         at System.Text.Json.Utf8JsonReader.TryGetNumber(ReadOnlySpan`1 data, Int32& consumed)
         at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
         at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
         at System.Text.Json.Utf8JsonReader.Read()
         at System.Text.Json.JsonDocument.TryParseValue(Utf8JsonReader& reader, JsonDocument& document, Boolean shouldThrow, Boolean useArrayPools)
         at System.Text.Json.JsonElement.ParseValue(Utf8JsonReader& reader)
         at HotChocolate.Types.JsonType.JsonFormatter.Format(IValueNode node)
         at HotChocolate.Types.JsonType.ParseLiteral(IValueNode valueSyntax)
         at HotChocolate.Types.InputParser.ParseLeaf(IValueNode resultValue, ILeafType type, Path path, IInputFieldInfo field)
         at HotChocolate.Types.InputParser.ParseLiteralInternal(IValueNode value, IType type, Path path, Int32 stack, Boolean defaults, IInputFieldInfo field)  
         at HotChocolate.Types.InputParser.ParseLiteralInternal(IValueNode value, IType type, Path path, Int32 stack, Boolean defaults, IInputFieldInfo field)  
         at HotChocolate.Types.InputParser.ParseObject(IValueNode resultValue, InputObjectType type, Path path, Int32 stack, Boolean defaults)
         at HotChocolate.Types.InputParser.ParseLiteralInternal(IValueNode value, IType type, Path path, Int32 stack, Boolean defaults, IInputFieldInfo field)  
         at HotChocolate.Types.InputParser.ParseLiteralInternal(IValueNode value, IType type, Path path, Int32 stack, Boolean defaults, IInputFieldInfo field)  
         at HotChocolate.Types.InputParser.ParseLiteral(IValueNode value, IInputFieldInfo field, Type targetType)
         at HotChocolate.Execution.Processing.MiddlewareContext.CoerceArgumentValue[T](ArgumentValue argument)
         at HotChocolate.Execution.Processing.MiddlewareContext.ArgumentValue[T](String name)
         at DataAnnotatedModelValidations.ValidatorMiddleware.<>c__DisplayClass6_0.<ReportErrorFactory>b__0(IInputField argument)
         at System.Linq.Parallel.ForAllOperator`1.ForAllEnumerator`1.MoveNext(TInput& currentElement, Int32& currentKey)
         at System.Linq.Parallel.ForAllSpoolingTask`2.SpoolingWork()
         at System.Linq.Parallel.SpoolingTaskBase.Work()
         at System.Linq.Parallel.QueryTask.BaseWork(Object unused)
         at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
      --- End of stack trace from previous location ---
         at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
         at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)

### Additional Context?

_No response_

### Version

13.0.5

onionhammer:
Fix issue parsing JSON scalar inputs

- Adds a 'bytes written' to call to `GetSpan` so that JsonElement.ParseValue only attempts to read the actual bytes written, rather than continuing to read the underlying buffer.
- Added test case

Closes #6023

### DIFF
--- a/src/HotChocolate/Core/src/Types/Types/Scalars/JsonType.cs
+++ b/src/HotChocolate/Core/src/Types/Types/Scalars/JsonType.cs
@@ -181,7 +181,7 @@ public sealed class JsonType : ScalarType<JsonElement>
             using var jsonWriter = new Utf8JsonWriter(bufferWriter);
             _visitor.Visit(node, new JsonFormatterContext(jsonWriter));
 
-            var jsonReader = new Utf8JsonReader(bufferWriter.GetSpan());
+            var jsonReader = new Utf8JsonReader(bufferWriter.GetSpan(jsonWriter.BytesPending));
             return JsonElement.ParseValue(ref jsonReader);
         }
 
--- a/src/HotChocolate/Core/src/Types/Types/Scalars/JsonType.cs
+++ b/src/HotChocolate/Core/src/Types/Types/Scalars/JsonType.cs
@@ -180,8 +180,9 @@ public sealed class JsonType : ScalarType<JsonElement>
             using var bufferWriter = new ArrayWriter();
             using var jsonWriter = new Utf8JsonWriter(bufferWriter);
             _visitor.Visit(node, new JsonFormatterContext(jsonWriter));
+            jsonWriter.Flush();
 
-            var jsonReader = new Utf8JsonReader(bufferWriter.GetSpan(jsonWriter.BytesPending));
+            var jsonReader = new Utf8JsonReader(bufferWriter.Body.Span);
             return JsonElement.ParseValue(ref jsonReader);
         }
 
--- a/src/HotChocolate/Core/test/Types.Mutations.Tests/AnnotationBasedMutations.cs
+++ b/src/HotChocolate/Core/test/Types.Mutations.Tests/AnnotationBasedMutations.cs
@@ -128,6 +128,27 @@ public class AnnotationBasedMutations
     }
 
     [Fact]
+    public async Task SimpleJsonMutationExtension_Inferred_Execute()
+    {
+        var result =
+            await new ServiceCollection()
+                .AddGraphQL()
+                .AddMutationType()
+                .AddTypeExtension<SimpleJsonMutationExtension>()
+                .AddMutationConventions(
+                    new MutationConventionOptions { ApplyToAllMutations = true })
+                .ModifyOptions(o => o.StrictValidation = false)
+                .ExecuteRequestAsync(
+                    @"mutation {
+                        doSomething(input: { something: 10 }) {
+                            string
+                        }
+                    }");
+
+        result.MatchSnapshot();
+    }
+
+    [Fact]
     public async Task Ensure_That_Directive_Middleware_Play_Nice()
     {
         var result =
@@ -981,6 +1002,13 @@ public class AnnotationBasedMutations
     {
         public string DoSomething(string something)
             => something;
+    }
+
+    [ExtendObjectType("Mutation")]
+    public class SimpleJsonMutationExtension
+    {
+        public string DoSomething(System.Text.Json.JsonElement something)
+            => "Done";
     }
 
     public class SimpleMutationAttribute
--- a/src/HotChocolate/Core/test/Types.Mutations.Tests/__snapshots__/AnnotationBasedMutations.SimpleJsonMutationExtension_Inferred_Execute.snap
+++ b/src/HotChocolate/Core/test/Types.Mutations.Tests/__snapshots__/AnnotationBasedMutations.SimpleJsonMutationExtension_Inferred_Execute.snap
@@ -1,0 +1,7 @@
+{
+  "data": {
+    "doSomething": {
+      "string": "Done"
+    }
+  }
+}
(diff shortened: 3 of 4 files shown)


michaelstaib:
Why not just flush?

src/HotChocolate/Core/src/Types/Types/Scalars/JsonType.cs
```diff
@@ -181,7 +181,7 @@ public static JsonElement Format(IValueNode node)
             using var jsonWriter = new Utf8JsonWriter(bufferWriter);
             _visitor.Visit(node, new JsonFormatterContext(jsonWriter));
```
michaelstaib, line 183: jsonWriter.Flush?

onionhammer: If you flush or close the writer, it advances the underlying buffer past the content

michaelstaib: Not following, the buffer is the internal memory buffer we use to write the json to. we can flush it an read it  then dispose it .... we anyway dispose it.

onionhammer: @michaelstaib It's difficult to explain, but I'll try to walk you through it.

My initial thought was that the buffer needed to be flushed, and that would solve it, but it turns out that was wrong. I will go through line-by-line a variation I tested thoroughly:

```C#
using var bufferWriter = new ArrayWriter();
using (var jsonWriter = new Utf8JsonWriter(bufferWriter))  // Using here closes and flushes the writer after visitor
    _visitor.Visit(node, new JsonFormatterContext(jsonWriter));

var jsonReader = new Utf8JsonReader(bufferWriter.GetSpan(bufferWriter.Length)); // The length here is now 2, however by disposing the jsonWriter, the `_start` variable on the bufferWriter is also 2 (since `Advance` was called), so the only remaining data in the buffer is all `0`'s.
return JsonElement.ParseValue(ref jsonReader);
```

image

Calling just .Flush has the same effect.

onionhammer: @michaelstaib here's an alternative:

image

michaelstaib: The analysis is wrong :) 

There was a bug ... and that was that I used the GetSpan(). 

GetSpan is a writer method to write into a segment of the writer ... in order to access the context you need to use `Body`. I fixed that and updated the PR. 

Body access the written data....

Your fix worked for this example but would have failed for larger jsons as you could just capture the pending bytes ... but when the writer does autoflush inbetween it would have been broken.

michaelstaib: We still need some proper tests for this here: `JsonTypeTests`

onionhammer: > Your fix worked for this example but would have failed for larger jsons as you could just capture the pending bytes ... but when the writer does autoflush inbetween it would have been broken.

Ahh yes, good point. I think your fix looks good though.

PR #6030, opened 2023-04-06
Fix issue converting idictionary to object #3617

### LINKED ISSUE #3617: AnyType IDictionary serialize like JSON
**Describe the bug**

`IDictionary<string, object>` serialization as `AnyType` is weird. 

**To Reproduce**

With the following query and output type

```cs
public class Query
{
    public Thing Query() => new Thing { Junk = new Dictionary<string, object> {{ "name", "yo" }}};
}

public class Thing {
  [GraphQLType(typeof(AnyType))] public IDictionary<string, object> Junk { get; set; }
}
```

An error is thrown `EXEC_INVALID_LEAF_VALUE` when querying.

If I change the dictionary type to `IDictionary<string, string>` then the serialization is weird

```gql
{
  "junk": [
    {
       "key": "name",
       "value": "yo"
    }
  ]
}
```

If I create a constructor to take the `dictionary<string, object>` and iterate over its values calling `ToString` then I get the expected result and that is weird.

```cs
foreach(var key in Junk.Keys) {
   Junk[key] = Junk[key].ToString();
}
```

But, I get the expected result

```gql
{
  "junk": {
    "name": "yo"
   }
}
```

**Expected behavior**

I expect `IDictionary<string, object>` to serialize like json

 - OS: macos
 - Version: 10.15.7

Spaier:
Allow to serialize IDictionary in AnyType.

Before `ObjectToDictionaryConverter` tried to serialize any `IDictionary` as an object instead of a dictionary when the type didn't implement `IReadOnlyDictionary`

STR:
1. Return IDictionary<string, object> from the query. Implementation of  IDictionary<string, object> shouldn't implement IReadOnlyDictionary<string, object>

Expected Result: `{ "foo": { "bar": 123 } }`
Actual Result: `{}`

Closes #3617

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
-    "sdk": {
-        "version": "6.0.100", 
-        "rollForward": "latestMinor"
-    }
+  "sdk": {
+      "version": "6.0.100",
+      "rollForward": "latestMinor"
+  }
 }
--- a/src/HotChocolate/Core/src/Types/Utilities/ObjectToDictionaryConverter.cs
+++ b/src/HotChocolate/Core/src/Types/Utilities/ObjectToDictionaryConverter.cs
@@ -90,8 +90,9 @@ internal class ObjectToDictionaryConverter
             var dict = new Dictionary<string, object>();
             setValue(dict);
 
-            if (obj is IReadOnlyDictionary<string, object> d)
+            if (obj is IReadOnlyDictionary<string, object> || obj is IDictionary<string, object>)
             {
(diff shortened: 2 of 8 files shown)


glen-84:
@michaelstaib Is this resolved by #6367?

glen-84:
Closing, as the tests are passing in the latest versions of Hot Chocolate. Likely resolved by #6367.

LorenzoMarchisio05 (commented):
shoulnt't the utility file serialize like a json also every type like Dictionary<string, int>, Dictionary<string, string>?

PR #6043, opened 2023-04-12
docs: Mention abstract classes can be used for union types

benmccallum:


benmccallum:
Hey @tobias-tengler, I think this is accurate, could probably even be based on a plain-old `class` (not abstract) but I'll let you decide if you think it's worth just saying "class" instead.

tobias-tengler (approved):
Sorry I missed this PR. Abstract class is fine.

PR #6054, opened 2023-04-15
Fixed DotNetVersion blank

### LINKED ISSUE #6048: The expression "[MSBuild]::VersionGreaterThanOrEquals('', 6)" cannot be evaluated. Version string was not in a correct format.

### Product

Strawberry Shake

### Describe the bug

I just finished the migration from v12 to v13 and found that I was having the same build problems mentioned in #6036.

After looking around some more I found the PR #6012 which tried to solve this problem; however, I'm running into other problems. Now when I build I receive the error

```
error MSB4184: The expression "[MSBuild]::VersionGreaterThanOrEquals('', 6)" cannot be evaluated. Version string was not in a correct format.
```

In addition to this there seems to be a problem with the `$(GraphQLCodeGenerationRoot)` property not setting properly in StrawberryShake.Server.targets line 23 because I'm seeing the C# Compiler error 

```
File name '\**\*.cs' is empty, contains invalid characters, has a drive specification without an absolute path, or is too long
```

### Steps to reproduce

1. While using a preview build of .NET 6 or 7, install version 14.0.0-tags-13-1-0-preview-5.1 of
   - StrawberryShake.Tools (as a dotnet tool)
   - StrawberryShake.Server (as a dependency)
 2. Try to build the project in VS 2022 17.6.0 Preview 3.0
 3. Try to build the project from the command line

### Relevant log output

```shell
C:\<redacted>\packages\strawberryshake.server\14.0.0-tags-13-1-0-preview-5.1\build\StrawberryShake.Server.targets(38,20): error MSB4184: The expression "[MSBuild]::VersionGreaterThanOrEquals('', 6)" c
annot be evaluated. Version string was not in a correct format. [C:\path\to\package\depending\on\strawberry\shake.csproj]
CSC : error CS2021: File name '\**\*.cs' is empty, contains invalid characters, has a drive specification without an absolute path, or is too long [C:\path\to\package\depending\on\strawberry\shake.csproj]
CSC : error CS2001: Source file '\**\*.cs' could not be found. [C:\path\to\package\depending\on\strawberry\shake.csproj]
```

### Additional Context?

The project I'm experiencing this issue with is targeting `net6.0`

When I run `dotnet --version` I get

```
7.0.300-preview.23179.2
```
I have tried to set the global.json to both 6.0.0 and 7.0.0 and still had the same issue.

I have tried building with Visual Studio 2022 17.6.0 Preview 3.0 and also from the dotnet CLI. 

🟢 There have been times when things seemed to be able to build when I manually set the `IntermediateOutputPath` in my .csproj file to

```xml
<IntermediateOutputPath>$(MSBuildProjectExtensionsPath)</IntermediateOutputPath>
```
or
```xml
<IntermediateOutputPath>$(MSBuildProjectExtensionsPath)\berry</IntermediateOutputPath>
``` 

but the ability to build seems fleeting and is not always reliable.

### Version

14.0.0-tags-13-1-0-preview-5.1

sunghwan2789:
This PR is a follow-up to #6012, which attempted to resolve a build issue with dotnet preview sdk.

- Extract `ExecTask` to a dependent target so that `ConsoleOutput` can be set to the `DotNetVersion` property before evaluating `PropertyGroup`.

Closes #6048

### DIFF
--- a/src/StrawberryShake/MetaPackages/Common/MSBuild/StrawberryShake.targets
+++ b/src/StrawberryShake/MetaPackages/Common/MSBuild/StrawberryShake.targets
@@ -24,16 +24,18 @@
     </ItemGroup>
   </Target>
 
-  <Target
-    Name="_GenerateGraphQLCode"
-    Inputs="@(GenerateGraphQLCodeItems)"
-    Outputs="$(MSBuildProjectDirectory)\$(IntermediateOutputPath)berry\.build.info"
-    Condition="'@(GraphQL)' != ''">
-
+  <Target Name="_DotNetVersion">
     <Exec Command="dotnet --version" StandardOutputImportance="Low" ConsoleToMsBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="DotNetVersion" />
     </Exec>
+  </Target>
 
+  <Target
+    Name="_GenerateGraphQLCode"
+    DependsOnTargets="_DotNetVersion"
+    Inputs="@(GenerateGraphQLCodeItems)"
+    Outputs="$(MSBuildProjectDirectory)\$(IntermediateOutputPath)berry\.build.info"
+    Condition="'@(GraphQL)' != ''">
     <PropertyGroup>
       <DotNetMajor Condition="$([MSBuild]::VersionGreaterThanOrEquals($(DotNetVersion), 6))">6</DotNetMajor>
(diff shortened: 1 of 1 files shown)


Psypher9:
Ah yes, this makes sense. I had tried a couple ways of trying to make sure the version would output but they didn't work. 

I can see how this would help

michaelstaib (approved):

PR #6067, opened 2023-04-18
Fix broken links

wilmerferreira:
This contains fixes for broken links in docs


tobias-tengler (changes requested):
Thanks for fixing the links, but the `<PackageInstallation ...`s you marked as code are actual React components being displayed on the Website. Marking them as code would be wrong. I understand that this is suboptimal for people viewing the raw markdown files. Do you know of any mechanism to fallback to text for raw markdown?

wilmerferreira:
Removed changes around React components

tobias-tengler (approved):

PR #6181, opened 2023-05-22
Authorize using roles amends

fiakkasa:
Amends to Authorization using roles


fiakkasa:
This will not be relevant anymore as the current behavior is more appropriate for the scenario where a use is authenticated but not authorized.

PR #6183, opened 2023-05-22
Adds documentation for environments

PascalSenn:


PascalSenn:
@rstaib  i would not version these docs. as this would require to change it every few weeks. 
The version is not really displayed anywhere right? shouldnt we just use it as is?

michaelstaib (approved):

PR #6512, opened 2023-09-07
Allow postgres subscriptions to be configured in publish only mode

### LINKED ISSUE #6511: Postgres Subscription Transport needs to be able to register the ITopicEventSender even when there is no graphql pipeline
### Product

Hot Chocolate

I am currently implementing subscriptions using the new Postgres backend. It works great!

However my solution has several components including an async job processor service. This service does not have an api and is event driven. It does not host a graphql pipeline.

But i still want to be able to dispatch events via the postgres backplane for the frontend api to pickup.

The service `ITopicEventSender` and its implementation `PostgresPubSub` is registered through the extension class `PostgresSubscriptionTransportExtensions.AddPostgresSubscriptions()`. `PostgresPubSub` is internal which means i can't register it in my backend service.

https://github.com/ChilliCream/graphql-platform/blob/49ade2117a1babed950d3e4b74abe8181b51ae77/src/HotChocolate/Core/src/Subscriptions.Postgres/PostgresSubscriptionTransportExtensions.cs#L39-L41

### The solution you'd like

I would like an additional extension method suitable for use in my backend service (based on `IServiceCollection` rather than `IRequestExecutorBuilder`) that registers `PostgresPubSub` as `ITopicEventSender` and the associated postgres connection requirements so i can dispatch events from the backend, through postgres to the front end api containers. 
Note: In this scenario the `ITopicEventReceiver` would not be required.

```cs

public static IServiceCollection AddPostgresSubscriptions(
        this IServiceCollection serviceCollection,
        Action<IServiceProvider, PostgresSubscriptionOptions> configure)

```

I can submit a PR for this if you like 👍

williamdenton:
Allows postgres subscription transport to be used in processes that perform publish operations only
see https://github.com/ChilliCream/graphql-platform/issues/6511

Closes #6511


williamdenton:
maybe too DRY? not sure if its best to duplicate some of the code rather than reuse it the way i have here

williamdenton:
I've just had another thought - the ITopicSender should participate in transactions (enlist) so the event is published when related data is committed to the database. Obviously the receiver needs to be independent of any transaction and pooling as it currently is

PascalSenn (approved):
This PR looks good. Thanks for the contribution!

(Just one minor formatting comment)

PascalSenn:
> I've just had another thought - the ITopicSender should participate in transactions (enlist) so the event is published when related data is committed to the database. Obviously the receiver needs to be independent of any transaction and pooling as it currently is

Yes i agree. this would solve a couple of issues. e.g:
- When you are in a transaction and you store an update and notify about the update, it's likely that the message is received on the consumer side before the data is actually updated in the database
- When your transaction fails, the notification still arrives. 

There are a couple of issue on the implemntation side with this though:
- We currenlty batch the send notifications for higher throughput on inserts (it's like 10x more throughput)
- The ITopicEventSenver runs decoupled from the context of the request

I have talked with @michaelstaib about this before, but we only really cut the topics ones in a discussion. Maybe we can pick this up again. There are a few adjustments that we would have to make to make this work.

williamdenton:
We discussed this internally too and it is tricky to reason about. Leaving it as is is simpler and we can easily work around it. Using a different backend like redis you would not get any guarantees on transaction related dispatch

michaelstaib (approved):

PascalSenn:
@williamdenton  Thanks for the contribution!

williamdenton:
I'm guessing this will be in 13.6 rather than 13.5.x?

PR #6688, opened 2023-11-10
Add IHttpConnectionFactory for non-HttpClient support

zaneclaes:
- Create `IHttpConnectionFactory` that returns `IHttpConnection`
- Implement `DefaultHttpConnectionFactory`
- Change code generation to use new factory

Closes https://github.com/ChilliCream/graphql-platform/issues/6647

I started by trying to modify `IHttpClientFactory`, but quickly realized it is deeply entangled with the Microsoft implementation thereof. Instead, I opted to implement a new factory for the higher-order `IHttpConnection`. However, this necessitated changing some dependencies, especially notably I had to add `Microsoft.Extensions.Http` to the StrawberryShake Tools package (assuming we want to colocate `DefaultHttpConnectionFactory` with `DefaultHttpClientFactory`, that is).

Finally, I'm not sure if I've correctly injected the DefaultHttpConnectionFactory into the DI. I would expect that StrawberryShake should automatically provide this feature for clients so that it does not break any existing clients when new code is generated. However, I'm not sure where that's currently being done for `DefaultHttpClientFactory` (I only see it referenced in CommandTools.cs, so I'm not sure how it's being injected by default for clients) @michaelstaib


zaneclaes:
In the last push, I was able to successfully test code generation. Unfortunately, I am struggling to run the tests locally. The codegen tests always fail, even on a fresh checkout (maybe I missed a step?) I believe because the `DEBUG` flag is not set in the project run configuration, meaning that the comments in the code are not generated and the output does not match the stubbed output:
Screenshot 2023-11-10 at 9 42 01 AM

zaneclaes:
Last thought: if preferred, we could omit the new classes and simply change the code generator to add the 2 lines at the top of the client generation function:
```
            global::Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddSingleton<global::StrawberryShake.IHttpConnection>(services, sp => {
                global::StrawberryShake.IHttpConnection? conn = parentServices.GetService<global::StrawberryShake.IHttpConnection>();
                if (conn != null) return conn;
                var clientFactory = global::Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService<global::System.Net.Http.IHttpClientFactory>(parentServices);
                return new global::StrawberryShake.Transport.Http.HttpConnection(() => clientFactory.CreateClient("TuneQuery"));
            });
```

Issue #3088, opened 2021-02-20
Changed argument name with NamingConventions is not found on mutations


nkovacic:
**Describe the bug**
When argument name in pure code first is changed with extending DefaultNamingConventions and GetArgumentMethod, it is not found on subsequent mutation execution, it returns that the old argument name is not found on the field

**To Reproduce**
1. Extend DefaultNamingConventions
````csharp
public class ExtendedNamingConventions: DefaultNamingConventions
    {
        public override NameString GetTypeName(Type type, TypeKind kind)
        {
            if (kind == TypeKind.InputObject)
            {
                var name = CleanInputEnding(type.Name);

                return new NameString(name);
            }
            
            else if (kind == TypeKind.Object)
            {
                var name = CleanEnding(type.Name);

                return new NameString(name);
            }

            return base.GetTypeName(type, kind);
        }

        public override NameString GetArgumentName(ParameterInfo parameter)
        {
            var nameString = base.GetArgumentName(parameter);

            if (nameString == null)
            {
                return nameString;
            }

            var name = CleanInputEnding(nameString.ToString());

            return new NameString(name.ToCamelCase());
        }

        private string CleanInputEnding(string name)
        {
            name = CleanEnding(name);

            if (!name.EndsWith("input", StringComparison.OrdinalIgnoreCase))
            {
                name += "Input";
            }

            return name;
        }

        private string CleanEnding(string name)
        {
            if (name.EndsWith("command", StringComparison.OrdinalIgnoreCase))
            {
                name = name.Replace("command", "", StringComparison.OrdinalIgnoreCase);
            }

            if (name.EndsWith("viewModel", StringComparison.OrdinalIgnoreCase))
            {
                name = name.Replace("viewModel", "", StringComparison.OrdinalIgnoreCase);
            }

            return name;
        }
    }
````
2. Create pure code first mutation extension
````csharp
[ExtendObjectType(nameof(Mutation))]
    public class IdentityMutation
    {
        public async Task<TokenResponseViewModel> Login(
            PasswordRequestTokenCommand passwordRequestTokenCommand, [Service] IMediator mediator, CancellationToken cancellationToken)
        {
            return await mediator.Send(passwordRequestTokenCommand, cancellationToken);
        }
}
````
3. Test in BCP
![image](https://user-images.githubusercontent.com/3670753/108597635-4a0d5180-738a-11eb-8e0b-decd7d898cac.png)

**Expected behavior**
[...]

PR #6705, opened 2023-11-17
Fixed argument naming conventions for mutation conventions.

PHILLIPS71:
- adds failing test to outline issue when enabling mutation conventions + custom naming conventions

Relates to [#3088](https://github.com/ChilliCream/graphql-platform/issues/3088)


glen-84:
Is this different from #6064?

PHILLIPS71:
> Is this different from #6064?

Yep, #6064 was in regards to the input and payload types not using the 'GetTypeName' in the 'DefaultNamingConvention'. 

I think it's due to the mutation name being picked up then being appending with Input or Payload, though unrelated to this issue.

Issue #5542, opened 2022-11-11
Owned entity returning null if any property is null in net7


magahl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Upgrading to net7 broke something regarding owned entities. Every time any property is null on the whole ntity return null. This does not happen in net6. 

The data is returned from the database, so its not that afaik. 

See results below. 
net6:
![image](https://user-images.githubusercontent.com/1409979/201289910-b21d8dee-f643-49df-868d-df75e46f92c0.png)

net7:
![image](https://user-images.githubusercontent.com/1409979/201289966-14f68e2e-b162-4771-9647-fd177f10d91d.png)

### Steps to reproduce

1.  See repo https://github.com/magahl/HotChoclateError

### Product

Hot Chocolate

### Version

13.0.0-preview.79

michaelstaib:
Re-verified on current `main` (February 25, 2026) with a focused EF owned-type repro using `UseProjection`: when one owned property is `null`, the owned object is still returned and only that field is `null`.\n\nI could not reproduce the original behavior on `net8`, `net9`, or `net10`. This appears fixed in current releases (v16/current main), so I am closing this issue.\n\nIf anyone can still reproduce this on a current version, please reopen with a minimal repro and exact package versions.

Issue #5229, opened 2022-07-11
Projection cause "Index out of range" on Database first EF core (net 6) 


invzz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I Have wrapped an old DB2 database with EF and created a DbContext which does not match the table schema on the database.
I had to "split" tables in different entities using the "Owned" properties 

so  namely the table `Viaggi` is splitted in `infoPax`, `infoMerci`,  and so on. 
for each of the owned type I created a Owned navigation property using EF

for example here entity is of Type `Viaggio`

```c#
 _ = entity.OwnsOne(via => via.InfoPax, entity =>
                {
                    _ = entity.Property(e => e.Capacita).HasColumnName("VIPAT").HasColumnType("smallint(2)").HasDefaultValueSql("0");
                    _ = entity.Property(e => e.Venduto).HasColumnName("VIPAV").HasColumnType("smallint(2)").HasDefaultValueSql("0");
                    _ = entity.Property(e => e.Warning).HasColumnName("VIPAW").HasColumnType("smallint(2)").HasDefaultValueSql("0");
                    _ = entity.Property(e => e.BloccoVendite).IsRequired().HasColumnName("VIPBV").HasColumnType("char(1)").HasDefaultValueSql("' '");
                    _ = entity.Property(e => e.Abilitazione).IsRequired().HasColumnName("VIPFV").HasColumnType("char(1)").HasDefaultValueSql("' '");
                    _ = entity.Property(e => e.TotaleMetriLieari).HasColumnName("VIPML").HasColumnType("decimal(5, 1)").HasDefaultValueSql("0");
                    _ = entity.Property(e => e.ProgressivoBooking).HasColumnName("VIPPB").HasColumnType("integer(4)").HasDefaultValueSql("0");
                }
```
I Have the usual repository accessing the data and returning an `IQueryable<Viaggio>`

If I don't use Projections everything works great. 

The problem starts when I add `[UseProjection]` on my query: It all works well until I try to access those owned (in theory non nullable) owned entities which really are in the same table of `Viaggi`.

That causes

```
"Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')",
```

I had a look in the produced Linq query :
```
 Compiling query expression: 
      'DbSet<Viaggio>()
          .Where(x => x.IdViaggio == __id_0)
          .Select(_s1 => new Viaggio{ 
              IdSocieta = _s1.IdSocieta, 
[...]

PR #6706, opened 2023-11-17
Avoid checking for null on non-nullable property for projection

### LINKED ISSUE #6604: Projection middleware checks for null on non-nullable field, resulting in exception for complex type in EF Core 8

### Product

Hot Chocolate

### Describe the bug

The projection middleware is generating this query:

```cs
DbSet<User>()
    .Where(u => u.Id == __query_EntityId_0)
    .Select(_s1 => new User{
        Id = _s1.Id,
        Username = _s1.Username,
        EmailAddress = _s1.EmailAddress,
        AccountStatus = _s1.AccountStatus != null ? new UserAccountStatus{ IsRegistrationCompleted = _s1.AccountStatus.IsRegistrationCompleted }
         : null
    }
    )
```

Which results in an EF Core 8 exception when using complex types:

> System.InvalidOperationException: Comparing complex types to null is not supported.

`User.AccountStatus` is non-nullable, so it shouldn't be checking for null like this.

### Steps to reproduce

I'll provide a reproduction repository if it helps, but it's probably not necessary.

### Relevant log output

_No response_

### Additional Context?

_No response_

### Version

13.5.1

glen-84:

- Check the nullability of properties before doing a null check.

Closes #6604.

---

- I'm not sure what `context.InMemory` is about.
- Is there an existing utility function/extension method for checking this nullability?
- Should these nullability checks be cached?

### DIFF
--- a/src/HotChocolate/Data/src/Data/Projections/Expressions/Handlers/QueryableProjectionFieldHandler.cs
+++ b/src/HotChocolate/Data/src/Data/Projections/Expressions/Handlers/QueryableProjectionFieldHandler.cs
@@ -14,6 +14,8 @@ public class QueryableProjectionFieldHandler
         selection.Field.Member is { } &&
         selection.SelectionSet is not null;
 
+    private readonly NullabilityInfoContext _nullabilityInfoContext = new();
+
     public override bool TryHandleEnter(
         QueryableProjectionContext context,
         ISelection selection,
@@ -90,7 +92,10 @@ public class QueryableProjectionFieldHandler
             return true;
         }
 
-        if (context.InMemory)
+        // TODO: Use existing utility function/extension method? Cache?
+        var nullabilityInfo = _nullabilityInfoContext.Create(propertyInfo);
+
+        if (context.InMemory && nullabilityInfo.ReadState == NullabilityState.Nullable)
         {
             parentScope.Level
                 .Peek()
--- a/src/HotChocolate/Data/test/Data.Projections.SqlServer.Tests/QueryableProjectionComplexTypeTests.cs
+++ b/src/HotChocolate/Data/test/Data.Projections.SqlServer.Tests/QueryableProjectionComplexTypeTests.cs
@@ -1,0 +1,63 @@
+using System.Threading.Tasks;
+using CookieCrumble;
+using HotChocolate.Execution;
+using Microsoft.EntityFrameworkCore;
+
+namespace HotChocolate.Data.Projections;
+
+#if NET8_0_OR_GREATER
+public class QueryableProjectionComplexTypeTests
+{
+    private static readonly Foo[] _fooEntities =
+    {
+        new() { Bar = new Bar { Baz = "testatest", } },
+        new() { Bar = new Bar { Baz = "testbtest", } },
+    };
+
+    private readonly SchemaCache _cache = new SchemaCache();
+
+    [Fact]
+    public async Task Create_Complex_Type()
+    {
+        // arrange
+        var tester = _cache.CreateSchema(_fooEntities, OnModelCreating);
+
+        // act
+        var res1 = await tester.ExecuteAsync(
+            QueryRequestBuilder.New()
+                .SetQuery(
+                    @"
+                        {
+                            root {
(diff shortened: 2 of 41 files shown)


glen-84:
@PascalSenn Is this feasible?

PascalSenn:
@glen-84  

We did remove it once for ef core and then added it again:
https://github.com/ChilliCream/graphql-platform/issues/6604#issuecomment-1810072311

So i am not sure what to do now. I believe if you do not do the nullcheck then it will include all the fields from the child or something like this.

glen-84:
You removed it completely, or conditionally based on the nullability of the property?

Should I be seeing something significant in the test snapshots, or is there no test that would highlight the previous issue? If not, then I think that we need that test to exist in order to have some more confidence here.

glen-84:
> So you only apply it when the property is nullable and ef does not complain?

By "does not complain", you mean that the tests in `HotChocolate.Data.Projections.SqlServer.Tests` pass? If so, yes, they all pass.

glen-84:
This will quite likely fix these issues as well:

- #5229
- #5542

src/HotChocolate/Data/src/Data/Projections/Expressions/Handlers/QueryableProjectionFieldHandler.cs
```diff
@@ -100,8 +102,9 @@ public override bool TryHandleLeave(
         }
 
         var memberInit = queryableScope.CreateMemberInit();
+        var nullabilityInfo = s_nullabilityInfoContext.Create(propertyInfo);
 
-        if (context.InMemory)
+        if (context.InMemory && nullabilityInfo.ReadState == NullabilityState.Nullable)
```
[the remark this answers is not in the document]
glen-84: Invalid according to Claude.

> Yes, that comment is largely invalid based on how NullabilityInfoContext actually works.
The concern it raises — that NullabilityInfoContext.Create may not provide accurate nullability information for value types — is unfounded. As covered above, value type nullability is the most reliable case for NullabilityInfoContext because it doesn't depend on nullable annotations or attributes at all. It's purely a CLR-level distinction: is the type Nullable<T> or not? The context handles this correctly and consistently.

src/HotChocolate/Data/src/Data/Projections/Expressions/Handlers/QueryableProjectionFieldHandler.cs
```diff
@@ -9,6 +9,8 @@ namespace HotChocolate.Data.Projections.Expressions.Handlers;
 public class QueryableProjectionFieldHandler
     : QueryableProjectionHandlerBase
 {
+    private readonly NullabilityInfoContext _nullabilityInfoContext = new();
```
michaelstaib: this is not thread-safe as far as i remember.

glen-84: Even when it's non-static?

glen-84: Okay, looking at solutions.

src/HotChocolate/Data/src/Data/Projections/Expressions/Handlers/QueryableProjectionFieldHandler.cs
```diff
@@ -102,7 +103,7 @@ public override bool TryHandleLeave(
         }
 
         var memberInit = queryableScope.CreateMemberInit();
-        var nullabilityInfo = _nullabilityInfoContext.Create(propertyInfo);
+        var nullabilityInfo = s_nullabilityInfoContext.Value!.Create(propertyInfo);
```
michaelstaib: I think you should not cache it at all ... cant we read the null information from the type system?

glen-84: You mean something like?:

```cs
field.Type.IsNullableType()
```

PR #6726, opened 2023-11-27
Collection expressions (IDE0300)

glen-84:

- Updated code triggering `IDE0300` to use collection expressions.

This handles all but 2 cases of `IDE0300`. The handled cases should be non-controversial (?). The remaining 2 cases are more subjective – do you have a preference for keeping the explicit types in the examples below?

```diff
-       _handlers = new ITypeRegistrarHandler[]
+       _handlers =
-       {
+       [
            new ExtendedTypeReferenceHandler(context.TypeInspector),
            new SchemaTypeReferenceHandler(),
            new SyntaxTypeReferenceHandler(context.TypeInspector),
            new FactoryTypeReferenceHandler(context),
            new DependantFactoryTypeReferenceHandler(context),
            new ExtendedTypeDirectiveReferenceHandler(context.TypeInspector)
-       };
+       ];
```

```diff
    public OffsetDateTimeType() : this(OffsetDateTimePattern.ExtendedIso)
    {
        // Backwards compatibility with the original code's behavior
        _serializationPattern = OffsetDateTimePattern.GeneralIso;
-       _allowedPatterns = new IPattern<OffsetDateTime>[] { OffsetDateTimePattern.ExtendedIso };
+       _allowedPatterns = [OffsetDateTimePattern.ExtendedIso];
    }
```

The pros of updating them:

- Terser code.
- Analyzer `IDE0300` can be fully enforced without exceptions.

The cons:

- Some developers may consider the explicit types to be more readable/understandable.

Since I could only find 2 examples within the entire codebase, it might make sense to just update these as well, and then enable the analyzer (and consider also using `EnforceCodeStyleInBuild`).


glen-84:
Superseded by #6845.

PR #6811, opened 2024-01-10
Fix spellcheck job

glen-84:

- Update Node.js for spellcheck job.
- Fix spelling errors.


tobias-tengler (approved):
Doesn't seem to run with the correct version in CI, but that's probably because the. change is coming from a fork?

glen-84:
> Doesn't seem to run with the correct version in CI, but that's probably because the. change is coming from a fork?

Ya, probably.

michaelstaib (approved):

PR #6867, opened 2024-02-08
fix: ApolloFederation does not require any entities to be present

dariuszkuc:
You can federate GraphQL services without specifying any entities (i.e. to expose all queries through the gateway). Current logic was incorrectly throwing exception where there were no entities.


dariuszkuc:
Opened up a follow up issue https://github.com/ChilliCream/graphql-platform/issues/6900 to track this.

PR #6898, opened 2024-02-14
Clarify middleware splitting vs. endpoints

cmeeren:
Addresses #6894

### DIFF
--- a/website/src/docs/hotchocolate/v12/security/authorization.md
+++ b/website/src/docs/hotchocolate/v12/security/authorization.md
@@ -362,7 +362,7 @@ public class Startup
 
 This method also accepts [roles](#roles) and [policies](#policies) as arguments, similar to the `Authorize` attribute / methods.
 
-> Warning: Unlike the `@authorize directive` this will return status code 401 and prevent unauthorized access to all middleware included in `MapGraphQL`. This includes our GraphQL IDE Banana Cake Pop. If we do not want to block unauthorized access to Banana Cake Pop, we can split up the `MapGraphQL` middleware and for example only apply the `RequireAuthorization` to the `MapGraphQLHttp` middleware.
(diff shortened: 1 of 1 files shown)


website/src/docs/hotchocolate/v12/security/authorization.md
```diff
@@ -362,7 +362,7 @@ public class Startup
 
 This method also accepts [roles](#roles) and [policies](#policies) as arguments, similar to the `Authorize` attribute / methods.
 
-> Warning: Unlike the `@authorize directive` this will return status code 401 and prevent unauthorized access to all middleware included in `MapGraphQL`. This includes our GraphQL IDE Banana Cake Pop. If we do not want to block unauthorized access to Banana Cake Pop, we can split up the `MapGraphQL` middleware and for example only apply the `RequireAuthorization` to the `MapGraphQLHttp` middleware.
+> Warning: Unlike the `@authorize directive` this will return status code 401 and prevent unauthorized access to all middleware included in `MapGraphQL`. This includes our GraphQL IDE Banana Cake Pop. If we do not want to block unauthorized access to Banana Cake Pop, we can split up the `MapGraphQL` middleware and for example only apply the `RequireAuthorization` to the `MapGraphQLHttp` middleware. Note however that when splitting up the middleware in this way, it is not possible to map the them to the same endpoint like `MapGraphQL` does.
```
glen-84, line 365:
```suggestion
> Warning: Unlike the `@authorize directive` this will return status code 401 and prevent unauthorized access to all middleware included in `MapGraphQL`. This includes our GraphQL IDE Banana Cake Pop. If we do not want to block unauthorized access to Banana Cake Pop, we can split up the `MapGraphQL` middleware and for example only apply the `RequireAuthorization` to the `MapGraphQLHttp` middleware. Note however that when splitting up the middleware in this way, it is not possible to map them to the same endpoint like `MapGraphQL` does.
```

PR #6916, opened 2024-02-18
Include path for Fusion subgraph errors

### LINKED ISSUE #6917: Fusion exposes subgraph errors without `path`
### Product

Hot Chocolate

### Version

14.0.0-p.36

https://github.com/ChilliCream/graphql-platform/pull/6916

### Steps to reproduce

Look at the tests linked in the PR above.

### What is expected?

I'd expect subgraph errors to be mapped into the fusion response with a `path`. The `path` should also be rewritten to the location in the fusion request document.

There is no `path` in the `errors` that get returned as part of the response from the Fusion server, which is [not spec compliant](https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format) and causes issues with error handling clients like Relay.

### Relevant log output

_No response_

### Additional context

_No response_

tobias-tengler:
Fixes #6917


tobias-tengler:
@michaelstaib I'd like some feedback on how I should best approach fixing this. I was thinking of including the errors returned from a subgraph on the `ExecutionState`. In the compose node I'd then check if there is an error associated with what I'm composing into the final result. I checked and I should have everything available to use the `PathHelper.CreatePathFromContext` to create a path that corresponds to the final result.
Do you already see a reason why that wouldn't work or why it's conceptually wrong? I'm still in the process of exploring how Fusion actually works...

PR #6918, opened 2024-02-18
Support ID list deserialization with custom Ids

### LINKED ISSUE #5114: Support Relay ID deserialization with custom Ids

faddiv:
Hi,
We ran into this problem when we tried to use plural custom id-s.

- CreateListFactory for custom types creates a string list instead of a custom type list. From this, the type converter can pick up it and create the correct type.

I used one of the recommendations from [this PR](https://github.com/ChilliCream/graphql-platform/pull/5160) to solve the issue.

Closes #5114

### DIFF
--- a/src/HotChocolate/Core/src/Types/Types/Relay/Serialization/CompositeNodeIdValueSerializer.cs
+++ b/src/HotChocolate/Core/src/Types/Types/Relay/Serialization/CompositeNodeIdValueSerializer.cs
@@ -14,7 +14,7 @@ namespace HotChocolate.Types.Relay;
 /// <typeparam name="T">
 /// The type of the value that is being serialized.
 /// </typeparam>
-public abstract class CompositeNodeIdValueSerialize<T> : INodeIdValueSerializer
+public abstract class CompositeNodeIdValueSerializer<T> : INodeIdValueSerializer
 {
     private const byte _partSeparator = (byte)':';
     private static readonly Encoding _utf8 = Encoding.UTF8;
--- a/src/HotChocolate/Core/test/Types.Tests/Types/Relay/DefaultNodeIdSerializerTests.cs
+++ b/src/HotChocolate/Core/test/Types.Tests/Types/Relay/DefaultNodeIdSerializerTests.cs
@@ -162,7 +162,7 @@ public class DefaultNodeIdSerializerTests
             [serializer]);
     }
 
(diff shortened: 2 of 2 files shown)


michaelstaib (changes requested):
Let me check why the type conversion does not work with typed lists ... I think the change should not be needed.

faddiv:
> Let me check why the type conversion does not work with typed lists ... I think the change should not be needed.

It doesn't work, because the createList creates a list with concrete type. In this case: new List<CustomId>()
but the id serializer serializes the CustomId with (d)efault type; when it deserializes it, string type is used. This results in error.
The type conversion happens after the ID deserialization, so that can't help. The old PR tried to bring in the conversion, but that would change the public API.

faddiv:
Hi Michael,
If you have time, please consider whether my implementation helps improve HotChocolate.
As far as I can tell, CustomId parsing in `List<>` is only possible this way since the conversion happens later.
I improved my solution. Now it works even if the serialized ID is not the default type.
I investigated the serializing side also, and there the non-default serialization only happens if a custom resolver is created, which transforms the value to one of the basic types.

faddiv:
Hi Michael,
As I see this is solved in this PR: https://github.com/ChilliCream/graphql-platform/pull/7026 so I closed.

PR #7016, opened 2024-03-26
Register interface types for operations in service collection

### LINKED ISSUE #6955: Strawberry Shake does not add interfaces of queries as services to DI container
### Product

Strawberry Shake

If you create any query, there is an interface and a class implementation generated for it. Only the direct class implementation is added to the service collection when using services.AddGraphQLClient(). This means the interface type cannot be used as a dependency on other classes.

A workaround is available, but this needs to be repeated manually for each query type:

`services.AddSingleton(sp => sp.GetRequiredService<IGraphQLClient>().GetTaskTypes);`

The interface type is needed so it can be mocked in unit tests.

### The solution you'd like

Register the generated query interfaces in the service collection.

abstractionnl:
Register interface types for operations in service collection

- Registers all operation interfaces as singleton for dependency injection.

Closes #6955


michaelstaib:
@abstractionnl can you sign the CLA? Otherwise we cannot move this thing forward?

glen-84 (approved):

glen-84:
@abstractionnl Thanks!

PR #7056, opened 2024-04-13
Fix incorrect cache control maxAge documentations

### LINKED ISSUE #7053: HotChocolate.Caching outputs max-age header in milliseconds instead of seconds
### Product

Hot Chocolate

### Version

13.8.1

https://github.com/kiangkuang/HotChocolateDemo

### Steps to reproduce

1. Clone this project that has been minimally configured like in https://www.youtube.com/watch?v=sk4XD3Zf64k
2. Run the project and execute the query
    ```
    query {
        book {
            title
        }
    }
    ```

### What is expected?

Response with header `Cache-Control: public, max-age=10`, because the units for `max-age` should be in seconds instead of milliseconds.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives
https://developers.cloudflare.com/cache/concepts/cache-control/

Response with header `Cache-Control: public, max-age=10000`, which was the milliseconds value specified in code.

### Relevant log output

_No response_

### Additional context

image
image

I'm not sure whether the better fix should either be to
1. Change the parameter units to be in Seconds or
2. Do the conversion from Milliseconds to Seconds when outputting the header, which has the ambiguity of whether CDNs or browsers expect non-integer values like `max-age=10.5` which arises due to the unit conversion.

kiangkuang:
Fixes incorrect caching XML documentations to align with header cache-control max-age usage and code behaviour of using seconds instead of milliseconds.

Closes #7053

### DIFF
--- a/src/HotChocolate/Caching/src/Caching/CacheControlAttribute.cs
+++ b/src/HotChocolate/Caching/src/Caching/CacheControlAttribute.cs
@@ -24,7 +24,7 @@ public sealed class CacheControlAttribute : DescriptorAttribute
     }
 
     /// <param name="maxAge">
-    /// The maximum time, in Milliseconds, the resource can be cached.
+    /// The maximum time, in seconds, the resource can be cached.
     /// </param>
     public CacheControlAttribute(int maxAge)
     {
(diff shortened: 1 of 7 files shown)


tobias-tengler (approved):
Thank you 🙏

tobias-tengler:
Can you maybe update ICacheControlOptions.DefaultMaxAge as well?

kiangkuang:
I've updated `ICacheControlOptions.DefaultMaxAge`, let me know if there's anywhere else I have missed.

PR #7130, opened 2024-05-29
Added directives to directive arguments

glen-84:

- Added directives to directive arguments.


glen-84:
This change seems to break:

`StrawberryShake/CodeGeneration/test/CodeGeneration.CSharp.Tests/SchemaGeneratorTests.cs#L28`

`MultipleInterfaceSchema.graphql` doesn't include the `listConstraint` directive. If I try to add it:

```graphql
directive @listConstraint(minItems: Int!) on ARGUMENT_DEFINITION
```

I get:

> 1. The specified directive `@listConstraint` is not allowed on the current location `ArgumentDefinition`. (HotChocolate.Types.DirectiveType)

**Edit:** It works if I add the directive to the end of the SDL ... but should this order matter?

PR #7142, opened 2024-06-04
feat: add char scalar to HotChocolate.Types.Scalars

PHILLIPS71:

- adds char scalar type to `HotChocolate.Types.Scalars`


michaelstaib:
We have decided that the use-cases are so small to have this in the library. Moreover, with version 16 its now very easy to create scalars. We leave this one to customer implementations.

PR #7249, opened 2024-07-09
Updated node ID serializers to allow internal IDs to be empty strings

### LINKED ISSUE #5090: Encoded ID with empty string value fails with ArgumentNullException

### Describe the bug

An empty string is a valid string. If the value part of a HotChocolate ID is of type string then it should serialize and deserialize without error.

For example, type name "MyType" with id value "" is base64-encoded as "MyType\nd" to "TXlUeXBlCmQ=". It base64-decodes back to "MyType\nd", 

Deserializing through HotChocolate.Types.Relay.IdSerializer causes an exception however:
```
Unhandled exception. System.ArgumentNullException: Array cannot be null. (Parameter 'bytes')
   at System.Text.Encoding.GetString(Byte* bytes, Int32 byteCount)
   at HotChocolate.Types.Relay.IdSerializer.CreateString(Span`1 serialized)
   at HotChocolate.Types.Relay.IdSerializer.Deserialize(String serializedId)
```

### Steps to reproduce

```
            IIdSerializer serializer = new IdSerializer(false);
            var ser = serializer.Serialize(new NameString("MySchema"), new NameString("MyType"), "");
            Console.WriteLine(ser);
            var deser = serializer.Deserialize(ser);
            Console.WriteLine(deser);
```

### Relevant log output

_No response_

### Additional Context?

_No response_

### Product

Hot Chocolate

### Version

12.9.0

glen-84:

- Updated node ID serializers to allow internal IDs to be empty strings.

Closes #5090

### DIFF
--- a/src/HotChocolate/Core/src/Types/Types/Relay/Serialization/OptimizedNodeIdSerializer.cs
+++ b/src/HotChocolate/Core/src/Types/Types/Relay/Serialization/OptimizedNodeIdSerializer.cs
@@ -216,14 +216,7 @@ internal sealed class OptimizedNodeIdSerializer : INodeIdSerializer
             throw SerializerMissing(typeName, rentedBuffer);
         }
 
-        var valueSpan = span.Slice(index + 1);
-        if (valueSpan.IsEmpty)
-        {
-            Clear(rentedBuffer);
-            throw new NodeIdInvalidFormatException(formattedId);
-        }
-
-        var value = ParseValue(valueSerializer, serializer.TypeName, valueSpan);
+        var value = ParseValue(valueSerializer, serializer.TypeName, span.Slice(index + 1));
         Clear(rentedBuffer);
         return value;
     }
--- a/src/HotChocolate/Core/src/Types/Types/Relay/Serialization/OptimizedNodeIdSerializer.cs
+++ b/src/HotChocolate/Core/src/Types/Types/Relay/Serialization/OptimizedNodeIdSerializer.cs
@@ -118,14 +118,7 @@ internal sealed class OptimizedNodeIdSerializer : INodeIdSerializer
             throw new NodeIdMissingSerializerException(typeNameString);
         }
 
-        var valueSpan = span.Slice(index + 1);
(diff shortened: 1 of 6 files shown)


michaelstaib (approved):

src/HotChocolate/Core/src/Types/Types/Relay/Serialization/OptimizedNodeIdSerializer.cs
```diff
@@ -223,14 +216,7 @@ public unsafe NodeId Parse(string formattedId, Type runtimeType)
...
 
-        var valueSpan = span.Slice(index + 1);
-        if (valueSpan.IsEmpty)
-        {
-            Clear(rentedBuffer);
-            throw new NodeIdInvalidFormatException(formattedId);
-        }
-
-        var value = ParseValue(valueSerializer, serializer.TypeName, valueSpan);
+        var value = ParseValue(valueSerializer, serializer.TypeName, span.Slice(index + 1));
```
michaelstaib, line 219: @glen-84 this was missing ... it has two parse methods.

src/HotChocolate/Core/src/Types/Types/Relay/Serialization/StringNodeIdValueSerializer.cs
```diff
@@ -34,6 +34,12 @@ public NodeIdFormatterResult Format(Span<byte> buffer, object value, out int wri
 
     public unsafe bool TryParse(ReadOnlySpan<byte> buffer, [NotNullWhen(true)] out object? value)
     {
+        if (buffer.Length == 0)
+        {
+            value = string.Empty;
```
michaelstaib, line 39: This is a bit nitpicking but ... I used the constant here instead of ""

src/HotChocolate/Core/src/Types/Types/Relay/Serialization/OptimizedNodeIdSerializer.cs
glen-84: D'Oh

PR #7253, opened 2024-07-10
Upgraded System.Text.Json to 8.0.4

tobias-tengler:
I was trying to reproduce an error locally when I suddenly could no longer build, since the System.Text.Json version was reported to have a critical security vulnerability. Seems to be [this one](https://github.com/advisories/GHSA-hh2w-p6rv-4g7w).
This PR bumps the package to the patched version and upgrades the minimum SDK version.


glen-84:
https://github.com/dotnet/runtime/issues/104619#issuecomment-2220381630

PR #7265, opened 2024-07-14
Updated more code to use C# collection expressions (IDE0300)

glen-84:

- Updated more code to use C# collection expressions (IDE0300).

---

ℹ️ I used the `when_types_exactly_match` option. The `when_types_loosely_match` option can change semantics (see [here](https://github.com/dotnet/roslyn/issues/72532#issuecomment-2185026497)).

When using the latter option:

```ini
dotnet_diagnostic.ide0300.severity = warning
dotnet_style_prefer_collection_expression = when_types_loosely_match
```

There are another 10-20 warnings. You might want to change all or some of these as well, but be careful regarding the change in semantics.


tobias-tengler:
Can't all of the byte arrays be changed to the new `"value"u8` instead?

glen-84:
@tobias-tengler https://github.com/ChilliCream/graphql-platform/pull/6591#issuecomment-1775267000

tobias-tengler:
Interesting. I wanted to know if the generated code is the same and found out that the UTF-8 string literals automatically add a null termination byte:
CleanShot 2024-07-14 at 19 16 10@2x
I would've thought it to be the same. The more you know :D

PR #7293, opened 2024-07-23
Return false when not found for TryGetValue

7rakir:
Return false when not found for `ObjectResult` `IReadOnlyDictionary.TryGetValue`

- `ObjectResult`'s explicit implementation of `IReadOnlyDictionary.TryGetValue` returns `true` when it should be `false`
- Adding a new set of tests in the image of `ObjectResultTests.TryGetValue_ValueIsFound`


michaelstaib (approved):
Good Catch! Thanks for contributing.

PR #7298, opened 2024-07-24
Removed `--property WarningLevel=0` in CI builds

glen-84:

- Removed `--property WarningLevel=0` in CI builds.

---

The build should fail until #7297 is merged.


glen-84:
There are some other warnings. Looking into it now.

PR #7312, opened 2024-07-28
Fixed SetPagingOptions

tobias-tengler:
During my refactor to `ModifyPagingOptions` I broke the now obsolete `SetPagingOptions`, by putting the options on the Schema context too early, i.e. before the `OnConfigureSchemaBuilderHooks` were ran.


src/HotChocolate/Core/test/Types.CursorPagination.Tests/IntegrationTests.cs
```diff
@@ -30,6 +30,24 @@ public async Task Simple_StringList_Schema()
...
     }
 
+    [Fact]
+    public async Task SetPagingOptionsIsStillApplied()
+    {
+        var executor =
+            await new ServiceCollection()
+                .AddGraphQL()
+                .AddQueryType<QueryType>()
+                .SetPagingOptions(new PagingOptions
```
glen-84, line 40: Can you suppress CS0618 here?

PR #7354, opened 2024-08-11
Improve activity reporting

### LINKED ISSUE #7320: Follow OTel conventions in default activity enricher
### Product

Hot Chocolate

HC's default activity enricher doesn't follow the OTel exception conventions at https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/ . See https://github.com/ChilliCream/graphql-platform/blob/a5bef9019004910e7bae8bd850b0980023ed176b/src/HotChocolate/Diagnostics/src/Diagnostics/ActivityEnricher.cs#L586-L612 .

### The solution you'd like

Could this be updated to report the exception directly using `activity.RecordException(error.Exception)`?

A workaround for now is:

```cs
private class OtelExceptionActivityEnricher : ActivityEnricher
{
    public OtelExceptionActivityEnricher(ObjectPool<StringBuilder> stringBuilderPoolPool, InstrumentationOptions options)
        : base(stringBuilderPoolPool, options) { }

    public override void EnrichResolverError(IMiddlewareContext context, IError error, Activity activity)
    {
        base.EnrichResolverError(context, error, activity);
        if (error.Exception != null)
            activity.RecordException(error.Exception);
    }
}
```

PascalSenn:
Closes #7320

### DIFF
--- a/src/HotChocolate/Diagnostics/src/Diagnostics/ActivityEnricher.cs
+++ b/src/HotChocolate/Diagnostics/src/Diagnostics/ActivityEnricher.cs
@@ -587,10 +587,15 @@ public class ActivityEnricher
 
     protected virtual void EnrichError(IError error, Activity activity)
     {
-        var tags = new ActivityTagsCollection()
+        if (error.Exception is { } exception)
+        {
+            activity.RecordException(exception);
+        }
+
+        var tags = new ActivityTagsCollection
         {
             new(AttributeExceptionMessage, error.Message),
-            new(AttributeExceptionType, error.Code),
+            new(AttributeExceptionType, error.Code)
         };
 
         if (error.Locations is { Count: > 0, })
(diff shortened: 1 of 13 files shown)


src/HotChocolate/Diagnostics/src/Diagnostics/ActivityEnricher.cs
```diff
@@ -582,33 +587,35 @@ void BuildPath()
...
+                activity.AddEvent(new ActivityEvent(AttributeExceptionEventName, default, tags));
             }
         }
-
-        activity.AddEvent(new("Error", tags: new(tags)));
+        else
+        {
+            activity.AddEvent(new ActivityEvent(AttributeExceptionEventName, default, tags));
+        }
     }
```
PascalSenn, line 615: @michaelstaib

If an `IError` has an exception, we now report the exception regularly. We also report the GraphQL Error as an "exception" to make it easier to trace them.

As the tags of an event should be static, we should not use `graphql.error.location[{i}].column`. With this PR, we currently report each location as its own error. It's unclear if this is the best approach, as it could lead to confusion when an error with multiple locations is reported as two separate errors.

Perhaps we should only report the first location to avoid this potential confusion.

michaelstaib (approved):

PR #7458, opened 2024-09-13
Document how to make @authorize directives public

Cyberboss:
I was trying to get @authorize directives to show up in my schema. Found a workaround, putting this in docs to hopefully save others the 3 hour headache.

Should they be internal?


Cyberboss:
This may not be all that's necessary. I'm still not seeing it attached to fields

Cyberboss:
Seems like visualizing directives like this isn't a widespread feature even with `EnableDirectiveIntrospection`. Also the `IntrospectionQuery` in bananacakepop doesn't support it either.

PR #7535, opened 2024-09-30
Removed Resolve overload that doesn't capture the result type

### LINKED ISSUE #6252: UseSorting on descriptors fail if the descriptor uses async

### Product

Hot Chocolate

### Describe the bug

I have a post in General on the slack channel where a workaround was provided
https://hotchocolategraphql.slack.com/archives/CD9TNKT8T/p1686204368383929

I get an error when i want to add sorting to a field using a descriptor. This only happens when its async

Example (this fails)
`descriptor.Field("networks").Type<NonNullType<ListType<NetworkType>>>().Resolve(async ctx =>
            {
                var dataloader = ctx.Service<INetworksForUserDataLoader>();
                var result = (await dataloader.LoadAsync(ctx.Parent<PlatformUser>().Id)).AsEnumerable();

				var checkSort = ctx.ArgumentKind("order") == ValueKind.Null ? Sort.None : Sort.Specified;
				if (checkSort == Sort.None)
					result = result.OrderByDescending(x => x.Name);

				return result;
            }).UseSorting();`
 
 `Does not work with UseSorting // Does work with useSorting
.Resolve(async ctx => // .Resolve( ctx =>`

Im not familiar with the internals but Anton on slack provided this feedback and an example of a quick fix
_I think it's because the async Resolve overload doesn't save the result type. In the implementation, they use NativeType, which is internal, to make it imply the type on its own. You can get it with reflection and pass it manually._

This is the error shown in visual studio
image

### Steps to reproduce

1. Have an descriptor on an entity which resolves using async.
2. Add UseSorting to the descriptor
3. Start api

### Relevant log output

_No response_

### Additional Context?

_No response_

### Version

13.2

glen-84:

- Removed `Resolve` overload that doesn't capture the result type.

Closes #6252


glen-84:
I think that this might also be causing this issue:

```cs
descriptor
    .Field("bookById")
    .Argument("id", a => a.Type<NonNullType<UuidType>>())
    .Resolve(async ctx => await ctx
        .Service<BookService>()
        .GetBookAsync(ctx.ArgumentValue<Guid>("id")));
```

> 1. Unable to infer or resolve the type of field Query.bookById. Try to explicitly provide the type like the following: `descriptor.Field("field").Type<List<StringType>>()`. (GettingStarted.Types.QueryType)

Using `Resolve<Book?>` works, but it should probably not be necessary?

glen-84:
Closing in favour of #9245.

PR #7559, opened 2024-10-03
fusion.md: Clarify when fusion or subgraph app id is required

bbrandt:
- Consistently specify `your-fusion-api-key` or `your-fusion-api-id` when the fusion config is required
- Consistently specify `your-subgraph-api-key` or `your-subgraph-api-id` when the subgraph config is required


PascalSenn (approved):
Very good addition! Thanks a lot!

PR #7589, opened 2024-10-11
Added predicates to DataLoader

PascalSenn:


src/GreenDonut/src/Core/DataLoaderFetchContext.cs
```diff
@@ -154,4 +155,30 @@ public ISelectorBuilder GetSelector()
         // a new default selector builder.
         return new DefaultSelectorBuilder();
     }
+
+    /// <summary>
+    /// Gets the predicate builder from the DataLoader state snapshot.
+    /// The state builder can be user to create a predicate expression.
```
glen-84, line 161:
```suggestion
    /// The state builder can be used to create a predicate expression.
```

src/GreenDonut/src/Core/DataLoaderFetchContext.cs
```diff
@@ -154,4 +155,30 @@ public ISelectorBuilder GetSelector()
...
+    /// </summary>
+    /// <returns>
+    /// Returns the predicate builder if it exists.
+    /// </returns>
+#if NET8_0_OR_GREATER
```
glen-84, line 166: These are not required on the `main` branch, but are required when cherry-picking into `main-version-14`.

glen-84 (changes requested):

PR #7765, opened 2024-11-25
Fix xml method description when dictionary args present

### LINKED ISSUE #7764: Unable to retrieve XML documentation for methods with `Dictionary` arguments
### Product

Hot Chocolate

### Version

14.0.0

https://github.com/smbecker/graphql-platform/blob/sb-xml-docs/src/HotChocolate/Core/test/Types.Tests/Types/Descriptors/Conventions/XmlDocumentationProviderTests.cs#L419

### Steps to reproduce

Include a method with a `Dictionary<string, string>` parameter.

```c#
public class WithDictionaryArgs
{
    /// <summary>
    /// This is a method description
    /// </summary>
    /// <param name="args">Args description</param>
    /// <returns></returns>
    public string Method(Dictionary<string, string>? args = null) => string.Empty;
}
```

### What is expected?

The documentation should be retrieved from `XmlDocumentationProvider` for the given method.

The XML path for methods with `Dictionary` arguments are generated incorrectly. The method name with a parameter of `Dictionary<string, string>` ends up with `Dictionary{System.String],[System.String}` rather than `Dictionary{System.String,System.String}`. This causes the XML element to not be found when calling `XmlDocumentationProvider.GetDescription` for the given method.

### Relevant log output

```shell

```

### Additional context

_No response_

smbecker:
Fixes issue where XML method descriptions could not be retrieved when the method included a `Dictionary` as one of the parameters.

Closes #7764

### DIFF
--- a/src/HotChocolate/Core/src/Types/Types/Descriptors/Conventions/XmlDocumentationProvider.cs
+++ b/src/HotChocolate/Core/src/Types/Types/Descriptors/Conventions/XmlDocumentationProvider.cs
@@ -405,7 +405,8 @@ public class XmlDocumentationProvider : IDocumentationProvider
                             "(`[0-9]+)|(, .*?PublicKeyToken=[0-9a-z]*)",
                             string.Empty)
                         .Replace("[[", "{")
-                        .Replace("]]", "}"))
+                        .Replace("]]", "}")
+                        .Replace("],[", ","))
                     .ToArray());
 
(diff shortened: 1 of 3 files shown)


michaelstaib (changes requested):
Your change has caused other tests you fail, can you have a look at the issues in the checks.

michaelstaib:
image

smbecker:
> image

Should be fixed

PR #7787, opened 2024-11-30
Added "AddDataLoader<TService, TImplementation>" executor overload that takes a factory

danielreynolds1:

- Adds an executor overload for dataloaders that allows an implementation factory to be provided


danielreynolds1:
- Should a separate PR be made into main?
- Should the executor extensions using the GreenDonut extension methods so code isn't being written twice?
  - The difference in GreenDonut is a call to TryAddDataLoaderCore which would be superfluous in this case

PR #7950, opened 2025-01-23
[Fusion] Add failing tests for selections below shared, non-refetchable field

tobias-tengler:


src/HotChocolate/Fusion/src/Core/Planning/Pipeline/ExecutionStepDiscoveryMiddleware.cs
```diff
@@ -797,9 +797,9 @@ private bool EnsureStepCanBeResolvedFromRoot(
 
         while (current is not null)
         {
-            var typeMetadata = _config.GetType<ObjectTypeMetadata>(path.Selection.DeclaringType.Name);
+            var typeMetadata = _config.GetType<ObjectTypeMetadata>(current.Selection.DeclaringType.Name);
 
-            if (!typeMetadata.Fields[path.Selection.Field.Name].Bindings.ContainsSubgraph(subgraphName))
+            if (!typeMetadata.Fields[current.Selection.Field.Name].Bindings.ContainsSubgraph(subgraphName))
```
tobias-tengler, line 802: This ensures that we fail the query plan generation early instead of either failing later or producing an invalid query plan that fails at runtime.

src/HotChocolate/Fusion/src/Core/Planning/Steps/SelectionPath.cs
```diff
@@ -35,8 +35,8 @@ public override string ToString()
 
         while (current is not null)
         {
-            sb.Append('/');
-            sb.Append(Selection.ResponseName);
+            sb.Insert(0, '/');
+            sb.Insert(1, current.Selection.ResponseName);
```
tobias-tengler, line 39: This is purely for debugging purposes. Previously the ToString would return something like `name/brand/productById` instead of `productById/brand/name` which would be more helpful to understand the path.

PR #8026, opened 2025-02-15
Fix ignored invalid field when enum is present

### LINKED ISSUE #8025: Invalid fields in variable ignored when an enum is present.
### Product

Hot Chocolate

### Version

15.0.3

https://github.com/faddiv/graphql-platform/tree/fix-ignored-field-when-enum-is-present

### Example model:

```csharp
public class User(string name)
{
    public string Name { get; set; } = name;
    public string? Surname { get; set; }
    public GenderEnum? Gender { get; set; }
}

public enum GenderEnum { Male, Female }

public class UserMutation
{
    public string AddUser(User user)
    {
        return $"{user.Name} {user.Surname} ({user.Gender}) was added!";
    }
}
```  

### Inputs

Execute the following GraphQL mutation:  

```graphql
mutation($user: UserInput!) { addUser(user: $user) }
```  

With the following variables:

```json
{
  "name": "Oliver",
  "surname": "Smith",
  "gender": "MALE",
  "foo": "bar"
}
```

### What is expected?

```json
{
  "errors": [
    {
      "message": "The field `foo` does not exist on the type `UserInput`.",
      "path": [
        "user"
      ],
      "extensions": {
        "type": "UserInput"
      }
    }
  ]
}
```

```json
{
  "data": {
    "addUser": "Oliver Smith (Male) was added!"
  }
}
```

### Relevant log output

```shell

```

### Additional context

In normal cases, if a variable contains an invalid field, the request returns an error stating that the field does not exist. However, if one of the variables is an **enum value**, then the invalid field is silently ignored instead of triggering an error.

The issue appears to be in the `VariableCoercionHelper` class, specifically in the `Rewrite` method. When it attempts to determine the field type and does not find it, the method **removes** the invalid field. As a result, the validation step does not detect the issue.

faddiv:
Fixes "Invalid fields in variable ignored when an enum is present"

- Extends `User`  in `VariableCoercionIntegrationTests` with an enum field and updates related fields.
- Adds an integration test to cover the case: `Invalid_Field_Provided_When_Enum_Is_Present`
- Adds an integration test to cover invalid enum value: `Invalid_Enum_Value_Provided`

Closes #8025

### DIFF
--- a/src/HotChocolate/Core/src/Execution/Processing/VariableCoercionHelper.cs
+++ b/src/HotChocolate/Core/src/Execution/Processing/VariableCoercionHelper.cs
@@ -173,6 +173,7 @@ internal sealed class VariableCoercionHelper
             {
                 // if we do not find a field on the type we also skip this error and let
                 // the deserialization produce a proper error on this.
+                fields?.Add(current);
                 continue;
             }
 
--- a/src/HotChocolate/Core/test/Execution.Tests/Integration/Components/VariableCoercionIntegrationTests.cs
+++ b/src/HotChocolate/Core/test/Execution.Tests/Integration/Components/VariableCoercionIntegrationTests.cs
@@ -151,13 +151,15 @@ public class VariableCoercionIntegrationTests
     }
 
     [Fact]
-    public async Task Invalid_Enum_Value_Provided()
+    public async Task Invalid_Field_Provided_When_Enum_Is_Present()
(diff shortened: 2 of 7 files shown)


michaelstaib (changes requested):
I think your fix brings unpredictability into the coercion. I will have a look at it today.

PR #8200, opened 2025-03-31
Unsubscribe from redis topic async

### LINKED ISSUE #8199: Unsubscribe from redis subscriptions asynchronously
### Product

Hot Chocolate

Currently the RedisTopic unsubcribes synchronously. This means that if you have a number of subscribtions completing at the same time, the applicaton threads are locked which can result in thread pool starvation.

### The solution you'd like

RedisTopic should unsubscribe asynchronously meaning the threads are freed up for other work.

dannyheard7:

Changes `DefaultTopic.OnConnectAsync` to return `IAsyncDisposable` instead of `IDisposable` (**breaking**) so that RedisTopic can unsubscribe asynchronously

Closes #8199


dannyheard7:
`CONTRIBUTING.md` mentions updating the unshipped public api text file, however I can't find this in the repository - if you let me know what needs updating instead then I'll do that

michaelstaib:
can you have a look at this PR and amend yours so that its not a breaking change?

https://github.com/ChilliCream/graphql-platform/pull/8202

dannyheard7:
I'm a bit thrown with that test failure, given the test output and it passes locally. I've loosen it up a bit to see if that helps.

Have you encountered anything similar before?

dannyheard7:
@michaelstaib  Any idea when we could see this merged in / released? We are seeing this cause performance issues in our server

michaelstaib:
@dannyheard7 the test still fails.

image

dannyheard7:
Ok i've removed the test as I can't think of a better way to test this and the mock seems flaky on the build server

michaelstaib:
Without test we cannot merge it ... we need to verify your change.

michaelstaib:
You have forked in a way that we cannot push changes into this PR ... we can take over the PR but than you loose your credits.

michaelstaib:
There are three things that need to be tested ... 

1. If you are not implementing `IAsyncDisposable` the standard dispose is triggered.
2. If you are implementing `IAsyncDisposable` the async dispose is triggered.

These two cover the change in the DefaultTopic.

3. You need to verify that the DisposeAsync works correctly on the RedisTopic you can also use squadron and there is no need to mock.

src/HotChocolate/Core/test/Subscriptions.Tests/DefaultTopicTests.cs
```diff
@@ -0,0 +1,74 @@
+﻿using HotChocolate.Subscriptions.Diagnostics;
+using HotChocolate.Tests;
+using Moq;
+using Xunit.Abstractions;
+
+namespace HotChocolate.Subscriptions;
+
+public class DefaultTopicTests(ITestOutputHelper outputHelper)
```
michaelstaib, line 8: This looks good

src/HotChocolate/Core/test/Subscriptions.Redis.Tests/RedisIntegrationTests.cs
```diff
@@ -47,6 +49,45 @@ public override Task Subscribe_And_Complete_Topic()
...
+
+        var activeChannelsAfterSubscribe = await GetActiveChannelsAsync();
+
+        await result.DisposeAsync();
+
+        var channelRemovedEvent = new ManualResetEventSlim(false);
+
+        _ = Task.Run(async () =>
+        {
+            while (true)
```
michaelstaib, line 70:
```suggestion
            while (cts.Token.IsCancellationRequested)
```

PR #8284, opened 2025-05-11
[Fusion] Abstract type & global object identification tests

tobias-tengler:


src/HotChocolate/Fusion-vnext/test/Fusion.Execution.Tests/Planning/FusionTestBase.cs
```diff
@@ -61,4 +65,51 @@ protected static void MatchInline(
         var actual = formatter.Format(plan);
         actual.MatchInlineSnapshot(expected + Environment.NewLine);
     }
+
+    protected record TestSubgraph([StringSyntax("graphql")] string Schema, bool Preprocess = true);
```
glen-84, line 69: Is it necessary to use this v1 preprocessing in tests?

I discussed the migration stuff with Michael today, and we settled on having a CLI command like `fusion migrate` to perform a once-off migration of a schema from v1 to v2, so this wouldn't happen during regular composition.

Thoughts?

PR #8292, opened 2025-05-17
fix: wrap missing DataLoader results in Result<T>.Resolve(default)

grounzero:
Wrap missing keys in BatchDataLoader and StatefulBatchDataLoader with Result<T>.Resolve(default) instead of null.


michaelstaib:
Can you explain why you want this?

PR #8363, opened 2025-06-16
Add variable filter tests

glen-84:
## Summary

- Adds tests for in-memory filtering over an expression-backed filter field (`latestItem`, bound to `Items.OrderByDescending(...).Take(1)`), verifying the query executes without errors whether the `where` argument is supplied inline or through a GraphQL variable.

## Test plan

- `Without_Variables`: `where` provided inline in the query, asserts no execution errors.
- `With_Variables`: the same filter provided via a `CollectionFilterInput!` variable, asserts no execution errors.


glen-84:
Superseded by #10126, which fixes the underlying variable coercion bug for expression-based filter fields and carries the regression test.

PR #8378, opened 2025-06-20
[Fusion] Fix bug with shared parent selection

tobias-tengler:
Given the following query:

```graphql
{
  something { # Subraph 1
    something  # Subraph 1
  }
  viewer {  # Subraph 1 & 2
    exclusiveSubgraphB  # Subraph 2
  }
}
```

The weighting algorithm will determine it should first plan for Subgraph 1.
The resulting execution step for Subgraph 1 will currently end up with the following selection set as it includes `viewer` but then doesn't have anything it can resolve from Subgraph 1:  
```graphql
{
  something {
    something
  }
  viewer {
   # empty
  }
}
```

This PR adds some checks to remove the empty `viewer` selection from the Subgraph 1 execution step, so it doesn't crash further down the line when attempting to create a request document for the subgraph call.


tobias-tengler:
Test failures seem unrelated.

PR #8383, opened 2025-06-25
[Fusion] Fix null being returned for batch item

tobias-tengler:


src/HotChocolate/Fusion/src/Core/Execution/Nodes/ResolveByKeyBatch.cs
```diff
@@ -339,44 +339,38 @@ private static JsonElement LiftData(JsonElement data, string[] path)
...
-                    key = ref Unsafe.Add(ref key, 1)!;
-                }
+                var lastKey = requires[^1];
+                var batchState = new BatchExecutionState(lastKey, state);
 
-                batchState = new BatchExecutionState(key, state);
+                batchExecutionState.Add(batchState);
 
                 state = ref Unsafe.Add(ref state, 1)!;
-                batchState = ref Unsafe.Add(ref batchState, 1)!;
```
tobias-tengler, line 379: This code path is never hit in the existing tests.
What's happening also doesn't make much sense to me so I cleaned it up.
- The StringBuilder was only ever being appended to, but the result wasn't used
- Incrementing the key to append to the StringBuilder would always end up with the last key being used to initialize the BatchState

src/HotChocolate/Fusion/src/Core/Execution/Nodes/ResolveByKeyBatch.cs
```diff
@@ -339,44 +339,38 @@ private static JsonElement LiftData(JsonElement data, string[] path)
...
-        ref var batchState = ref MemoryMarshal.GetArrayDataReference(batchExecutionState);
         ref var end = ref Unsafe.Add(ref state, executionState.Count);
 
         if (requires.Length == 1)
         {
             while (Unsafe.IsAddressLessThan(ref state, ref end))
             {
-                var key = FormatKeyValue(state.VariableValues[requires[0]]);
-                batchState = new BatchExecutionState(key, state);
+                if (state.VariableValues.ContainsKey(requires[0]))
```
tobias-tengler, line 353: This check is the relevant change

PR #8408, opened 2025-07-01
Added switch to disable the LocalDate/LocalTime/Date format check

glen-84:

- Added switch to disable the LocalDate/LocalTime/Date format check.


jorrit:
@glen-84 How do I configure my application to use this new switch?

glen-84:
```cs
builder.AddType(new LocalDateType(disableFormatCheck: true));
```

Note: This is currently only available in 16.x.

jorrit:
Thanks!

PR #8425, opened 2025-07-06
Added support for reference resolvers on an entity interface.

### LINKED ISSUE #7699: No way to add reference resolver to interface
### Product

Hot Chocolate

### Version

13.6.1

https://github.com/ChilliCream/graphql-platform/blob/984e263eb5f71d81c935d343e043f186543ae8e1/src/HotChocolate/Core/src/Types/Types/Descriptors/Contracts/IInterfaceTypeDescriptor.cs#L6

### Steps to reproduce

Create a new `InterfaceType<>` class and create the `Configure` method. On `IInterfaceTypeDescriptor`, there is no `Key()` method to add a key and reference resolver to the interface like there is on `IObjectTypeDescriptor`.

### What is expected?

Because `@interfaceObject` is supported, I would expect there to be the `Key()` method to add the key and a reference resolver (which is required to use  `@interfaceObject`).

No way to add a reference resolver, resulting in an error from Apollo when attempting to resolve the interface. This would be a feature request if not for the fact that `@interfaceObject` is supposed to be supported. But maybe there is another way to register a reference resolver to the interface that isn't documented.

### Relevant log output

_No response_

### Additional context

_No response_

danielreynolds1:

- Added support for reference resolvers on an entity interface.

Closes #7699


danielreynolds1:
It isn't entirely clear if the interface should interact with the "_Entity" field. When enabling this, you end up with an error that a union can't contain an interface.

Issue #6727, opened 2023-11-27
Fusion tries to resolve data from subgraph without top-level resolver


interad-woergoetter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Product

Hot Chocolate

### Describe the bug

Fusion will try to resolve data from the first subgraph that has all (or the most of) the requested data in its schema, regardless of whether that subgraph has a top-level resolver for the type in question.
This leads to an error during execution plan creation.

It appears that the query plan generator iterates over all subgraphs in order, i.e. alphabetically, meaning the names of the subgraphs impact the reproducability of this issue.

### Steps to reproduce

1. Set up a Fusion gateway server with the attached gateway package. 
[gateway.fgp.zip](https://github.com/ChilliCream/graphql-platform/files/13474261/gateway.fgp.zip)
2. Start the gateway server.
3. Send the query `query { foos { id name nameReversed } }` to the gateway.
4. Confirm that the gateway is able to create a query plan fetching additional data from Subgraph3.
5. Now send the query `query { foos { id name } }` to the gateway.
6. Notice that the gateway is not able to create a query plan.

### Relevant log output

```shell
{
  "errors": [
    {
      "message": "Unexpected Execution Error",
      "extensions": {
        "message": "There must be a field resolver definition in this context!",
[...]

PR #8497, opened 2025-08-06
[Fusion] Check if type can be resolved when selecting best matching subgraph

dawust:
We've run into the same issue as https://github.com/ChilliCream/graphql-platform/issues/6727
Feel free to take a look, feedback is welcome! Thanks!

We check if the schema contains the necessary resolver for a type when selecting the best matching subgraph


tobias-tengler:
Hey Dani :) Thanks for the contribution!

The change seems sensible to me, but it appears some tests in `src/HotChocolate/Fusion/test/Core.Tests` are now stuck in an endless loop. Could you check that?

Also it would be nice, if you could add a test like [this](https://github.com/ChilliCream/graphql-platform/blob/fdd5af3966d256addba31bc79ed374b95607496f/src/HotChocolate/Fusion/test/Core.Tests/DemoIntegrationTests.cs#L18C5-L93C6) that showcases the effect of your fix.

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
```diff
@@ -58,6 +61,13 @@ private static int EvaluateSubgraphCompatibilityScore(
                 {
                     score++;
```
tobias-tengler, line 75: I think this should no longer be here, right?

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
```diff
@@ -58,6 +61,13 @@ private static int EvaluateSubgraphCompatibilityScore(
...
 
+                    if (parentSelectionPath is null ||
+                        typeMetadataContext.Resolvers.ContainsResolvers(schemaName) ||
+                        configuration.EnsureStepCanBeResolvedFromRoot(schemaName, parentSelectionPath))
```
tobias-tengler: I think we can combine this with the if above that checks for introspection fields?
If we can't reach the current selection we shouldn't increment the score and there's also no point in visiting the child selections.

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
dawust: and it's gone

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
dawust: Merged the statements and no subgraph schema is returned anymore if we can't reach the selection. Throw_Instead_Of_Creating_Faulty_QueryPlan ended up looping all the time, if the wrong subgraph was chosen.

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
```diff
@@ -50,11 +55,24 @@ private static int EvaluateSubgraphCompatibilityScore(
         {
             var (currentSelections, currentTypeContext) = stack.Pop();
 
+            if (currentSelections.Count == 0)
```
tobias-tengler, line 61: What is this here for? Could you maybe add an explanatory comment?

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
```diff
@@ -50,11 +55,24 @@ private static int EvaluateSubgraphCompatibilityScore(
...
+                    currentTypeContext
+                        .Fields[selection.Field.Name]
+                        .Bindings
+                        .ContainsSubgraph(schemaName) &&
+                    (pathCanBeResolvedFromRoot ||
+                        currentTypeContext
+                            .Fields[selection.Field.Name]
+                            .Resolvers
+                            .ContainsResolvers(schemaName) ||
+                        typeMetadataContext.Resolvers.ContainsResolvers(schemaName)))
```
tobias-tengler: I think checking the new condition for child selections is wrong. Consider the following scenario

```graphql
# Subgraph A
type Query {
  productBySlug: Product
}

type Product {
  id: ID!
}
```

```graphql
# Subgraph B
type Query {
  productById(id: ID!): Product
}

type Product {
  id: ID!
  author: Author
}

type Author {
  name: String!
}
```

```graphql
# Subgraph C
type Query {
  productById(id: ID!): Product
}

type Product {
  id: ID!
  author: Author
}

type Author {
  name: String!
  rating: Int!
}
```

Planning a query like this:

```graphql
{
  productBySlug {
    author {
      name
      rating
    }
  }
}
```

The new condition will be false for name and rating on author, so both B & C end up with the same score, even though C could resolve more selections.

I think we should split up the method to only evaluate wether the type or path can be resolved for the "root" selections passed into `GetBestMatchingSubgraph`, but not their child selections.

Could you maybe create a test for the scenario I described to see if I'm wrong or if this could actually be a problem? Thanks :)

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
dawust: done

src/HotChocolate/Fusion/src/Core/Planning/QueryPlannerHelpers.cs
dawust: Added a test for your scenario. Subgraph C is chosen correctly :)

I also created a method which checks if the type or path can be resolved for the "root" selections as you mentioned. However considering the Query_Plan_36 and Query_Plan_37 tests, we still have to check if the currentTypeContext contains a resolver for the current selection field.

PR #8498, opened 2025-08-06
[Fusion] Ensure at least one Query.node lookup exists for a type implementing the Node interface

tobias-tengler:
We decided to have a fixed fallback query for node field execution steps like:

```graphql
query fallbackQuery($id: ID!) {
  node(id: $id) {
    # node selections
  }
}
```

That way we don't have a type explosion, since we only have to plan for types that actually have a type refinement below the `node` field.

For this to work each type implementing `Node` must be accessible through a `Query.node` lookup in a source schema.
This PR adds that validation.


src/HotChocolate/Fusion-vnext/test/Fusion.Composition.Tests/SatisfiabilityValidatorTests.cs
```diff
@@ -2028,18 +2029,33 @@ type Cat implements Node {
                         id: ID!
                         name: String!
                     }
+                    """,
+                    """
+                    # Schema B
+                    type Query {
+                      node(id: ID!): Node @lookup
```
glen-84: Indentation.

glen-84 (changes requested):

src/HotChocolate/Fusion-vnext/test/Fusion.Composition.Tests/SatisfiabilityValidatorTests.cs
tobias-tengler:
```suggestion
                        node(id: ID!): Node @lookup
```

glen-84 (approved):

PR #8561, opened 2025-08-18
[Fusion] Properly format multiline compose output

tobias-tengler:
I found long lists of errors a bit hard to scan, since the hierarchy was slightly interrupted by the second line not being aligned below the first line's text.
Thought about also covering the `[ERR] ` with indentation, but it got a bit too spaced out for my taste...

**Before**
CleanShot 2025-08-18 at 13 55 07@2x

**After**
CleanShot 2025-08-18 at 14 58 06@2x

### DIFF
--- a/src/HotChocolate/Fusion-vnext/src/Fusion.CommandLine/Commands/ComposeCommand.cs
+++ b/src/HotChocolate/Fusion-vnext/src/Fusion.CommandLine/Commands/ComposeCommand.cs
@@ -616,7 +616,7 @@ internal sealed class ComposeCommand : Command
 
     /// <summary>
     /// Since we're prefixing the message with an emoji and space before printing,
-    /// we need to also indent each line of a multiline message by two spaces to fix the alignment.
+    /// we need to also indent each line of a multiline message by three spaces to fix the alignment.
     /// </summary>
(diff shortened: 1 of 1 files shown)


glen-84:
Hmm. Why is it only 2 spaces and not 5?

```
❌·[ERR] ...
·····Test
```

2 for the emoji, 1 for the space, and 2 for indentation.

Edit: Oh it's already indented ... 3 then, not 5. 🙂

src/HotChocolate/Fusion-vnext/src/Fusion.CommandLine/Commands/ComposeCommand.cs
```diff
@@ -613,4 +613,20 @@ static async Task ReadSourceSchemaAsync(
             sourceSchemas.TryAdd(schemaName, (new SourceSchemaText(schemaName, sourceText), settings));
         }
     }
+
+    /// <summary>
+    /// Since we're prefixing the message with an emoji and space before printing,
+    /// we need to also indent each line of a multiline message by two spaces to fix the alignment.
```
glen-84:
```suggestion
    /// we need to also indent each line of a multiline message by three spaces to fix the alignment.
```

PR #8565, opened 2025-08-18
[Fusion] Add support for shared paths to satisfiability validation

tobias-tengler:


src/HotChocolate/Fusion-vnext/src/Fusion.Composition/Extensions/MutableObjectTypeDefinitionExtensions.cs
```diff
@@ -19,6 +19,12 @@ public static bool ExistsInSchema(this MutableObjectTypeDefinition type, string
             d => d.Name == FusionType && (string)d.Arguments[Schema].Value! == schemaName);
     }
 
+    public static bool ExistsInSchema(this MutableOutputFieldDefinition field, string schemaName)
```
glen-84: This should be in `MutableOutputFieldDefinitionExtensions`.

PR #8650, opened 2025-09-02
Docs: Added examples for With-methods

N-Olbert:

- Docs: Added examples for With-methods
- Test: Added validation for disposed HttpClient

Related
- https://github.com/ChilliCream/graphql-platform/issues/3467#issuecomment-2900815908
- https://hotchocolategraphql.slack.com/archives/CMUJ40EHJ/p1756738589396969?thread_ts=1756722636.163949&cid=CMUJ40EHJ

### DIFF
--- a/website/src/docs/strawberryshake/v16/networking/authentication.md
+++ b/website/src/docs/strawberryshake/v16/networking/authentication.md
@@ -7,12 +7,14 @@ Each network protocol of Strawberry Shake handles authentication a bit different
 
 # HTTP
 
-Strawberry Shake uses the `HttpClientFactory` to generate a `HttpClient` on every request.
-You can either register a `HttpClient` directly on the `ServiceCollection` or use the `ConfigureHttpClient` method on the client builder.
+There are several ways to adjust how Strawberry Shake obtains the `HttpClient` that will be used to perform HTTP calls.
+Unless specified otherwise, Strawberry Shake uses the `HttpClientFactory` to generate a new `HttpClient` for every request.
 
 ## ConfigureHttpClient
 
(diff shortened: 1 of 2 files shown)


website/src/docs/strawberryshake/v16/networking/authentication.md
```diff
@@ -61,6 +66,23 @@ services.AddHttpClient(
...
+```csharp
+IConferenceClient client = services.GetRequiredService<IConferenceClient>();
+
+var result = await client.GetSessions
+    .WithRequestUri(new Uri("https://localhost:5001/graphql"))
+    .WithHttpClient(new MyCustomHttpClient())
+    .ExecuteAsync();
+```
+
+> Note: The provided `HttpClient` will be disposed once `ExecuteAsync` completes.
```
N-Olbert, line 83: I emphasized this since it was very common in .NET Framework 4 to have one singleton HttpClient-instance that is used for each and every request.
This is also still the recommended way [(docs)](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use) if one doesn`t use dependency injection / IHttpClientFactory.

With this in mind, something like
```
private static readonly HttpClient httpClient = new HttpClient();


var result1 = await client.GetSomething.WithHttpClient(httpClient).ExecuteAsync();
var result2 = await client.GetSomethingElse.WithHttpClient(httpClient).ExecuteAsync();
...
```
will break.

I wonder whether it might make sense to add a boolean to control whether the client should be disposed to `WithHttpClient`  (I could add that) @michaelstaib

PR #8675, opened 2025-09-05
Add OnDataLoaded EventCallback

### LINKED ISSUE #8673: OnDataLoad Event on UseQuery blazor
### Product

Strawberry Shake

I am using the useQuery component in my Blazor application to fetch data. Currently, the component lacks a dedicated event that triggers when the data is successfully loaded. This makes it challenging to perform actions (e.g.,  triggering side effects) immediately after the data fetch completes.

### The solution you'd like

The useQuery component should provide an OnDataLoaded event or callback that triggers when the data is successfully fetched and available.

This event should pass the fetched data as a parameter, allowing developers to handle it in their application logic.

behroozbc:

- Add OnDataLoaded EventCallback on UseQueryCompoent of blazor

I couldn’t test this because every time I tried to add it to my Blazor project, I got other errors that weren’t related at all.
But based on my experience, I think this is correct.
Please test it if possible.
Also, I couldn’t find UseQuery component tests and didn’t write them. If needed, please guide me.
Additionally, I have a complaint about CONTRIBUTING.md because the information is outdated!
Closes #8673

### DIFF
--- a/src/StrawberryShake/Client/src/Razor/UseQuery.cs
+++ b/src/StrawberryShake/Client/src/Razor/UseQuery.cs
@@ -21,6 +21,8 @@ public abstract class UseQuery<TResult> : ComponentBase, IDisposable where TResu
 
     [Parameter] public RenderFragment? LoadingContent { get; set; }
 
+    [Parameter] public EventCallback<TResult> OnDataLoaded { get; set; }
+
     protected void Subscribe(IObservable<IOperationResult<TResult>> observable)
     {
         _subscription?.Dispose();
@@ -52,6 +54,7 @@ public abstract class UseQuery<TResult> : ComponentBase, IDisposable where TResu
         if (_isSuccessResult)
         {
             builder.AddContent(0, ChildContent, _result!);
+            OnDataLoaded.InvokeAsync(_result);
         }
 
         base.BuildRenderTree(builder);


behroozbc:
I see @glen-84 you want to review this pull requests, how I can help you to this PR be merge sooner?

michaelstaib:
We have put the PR on our internal backlog for discussion @glen-84 will review the actual issue and then we will see what design will fit to your issue.

src/StrawberryShake/Client/src/Razor/UseQuery.cs
```diff
@@ -21,6 +21,8 @@ public abstract class UseQuery<TResult> : ComponentBase, IDisposable where TResu
 
     [Parameter] public RenderFragment? LoadingContent { get; set; }
 
+    [Parameter] public EventCallback<TResult> OnDataLoaded { get; set; }
```
glen-84:
```suggestion
    [Parameter] public EventCallback<IOperationResult<TResult>> OnOperationResult { get; set; }
```

This will provide access to other properties of the result, like `Errors`, `IsErrorResult`, and `IsSuccessResult`.

src/StrawberryShake/Client/src/Razor/UseQuery.cs
```diff
@@ -52,6 +54,7 @@ protected override void BuildRenderTree(RenderTreeBuilder builder)
         if (_isSuccessResult)
         {
             builder.AddContent(0, ChildContent, _result!);
+            OnDataLoaded.InvokeAsync(_result);
```
glen-84: I would assume that this should be invoked in the lambda passed to `observable.Subscribe` in the `Subscribe` method?

src/StrawberryShake/Client/src/Razor/UseQuery.cs
glen-84, line 1: We should add this to the `UseSubscription` class as well.

glen-84 (changes requested):

behroozbc:
I think this have be on another pull request.
I saw the use query and use subscription have many same things, I like to know why these two classes cannot inheritance from same class, for example componentbase-> UseBase class( this name must change)-> usequery, usesubscription.

glen-84:
LGTM, I'm just waiting for @michaelstaib to comment on the testing aspect.

michaelstaib:
I will create a preview this week and ping you once its available.

PR #8684, opened 2025-09-08
Added global ID formatter conditionally

### LINKED ISSUE #7855: Using IdOperationFilterInputType as field filter type gives error when filtering
### Product

Hot Chocolate

### Version

14.3.0

https://github.com/dalen/hotchocolate-reproduction/tree/main/HotChocolateIdFilter

### Steps to reproduce

On a filter input type set the filter for a field to `.Type<IdOperationFilterInputType>()` to make the filter use ID for the field instead of something like `Int`. Then try to query it.

```
{
  books(where:  {
     id:  {
        eq: 1
     }
  }) {
    nodes {
      id
      title
    }
  }
}
```

Response

```json
{
  "errors": [
    {
      "message": "The node ID string has an invalid format.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "books"
      ],
      "extensions": {
        "originalValue": "1"
      }
    }
  ],
  "data": {
    "books": null
  }
}
```

### What is expected?

Expect the filtering to work and convert the ID to the underlying type, just like if you set an attribute to use `IdType`.

Instead I get the above error.

### Relevant log output

```shell

```

### Additional context

Using `id:  { eq: "1"}` with the ID as a string makes no difference.

glen-84:

- Added global ID formatter conditionally.

Closes #7855


tobias-tengler:
Could you add this to the breaking change docs as well. I think this could break a bunch of stuff. I know of at least one instance in my old company where we have a subgraph that doesn't register the Node interface, but accepts a Base64 encocded Id that should be decoded.

glen-84:
@tobias-tengler Reminds me of https://xkcd.com/1172/. 😆

This will mean that we can't cherry-pick it into 15?

PR #8716, opened 2025-09-16
Projections: Support selecting only __typename on union types

### LINKED ISSUE #8252: Regression bug: Unable to spread fragment on inheritance hierarchy added as a union or interface type when deferred to Entity Framework
### Product

Hot Chocolate

### Version

15

https://github.com/ChilliCream/graphql-platform/issues/4318

### Steps to reproduce

Looks like this issue is back in HC 15:
https://github.com/ChilliCream/graphql-platform/issues/4318

### What is expected?

See:
https://github.com/ChilliCream/graphql-platform/issues/4318

See:
https://github.com/ChilliCream/graphql-platform/issues/4318

### Relevant log output

```shell

```

### Additional context

_No response_

N-Olbert:

- Support selecting only __typename (without any other 'real' members) on union type

Closes #8252  (in this specific format)

Background:

If no 'real' member of an entity was requested as part of the selection set, like in

```graphql
{
  foo {
    items { # This is a union of ConcreteType1|ConcreteType2
      __typename 
      ... on ConcreteType2 {
          foo
       }
    }
  }
}
```

the projected query currenty looks like this:

```csharp
repository.Select(_s1 => (_s1 is ConcreteType1) ? new ConcreteType1 {}
                : (_s1 is ConcreteType2) ? (ConcreteType2)new ConcreteType2 { Foo = ((ConcreteType2)_s1).Foo }
                : default(BaseType));
```

However, `new ConcreteType1 {}` cannot be translated to SQL, which leads to the following exception:

>The client projection contains a reference to a constant expression of 'ConcreteType1'. This could potentially cause a memory leak; consider assigning this constant to a local variable and using the variable in the query instead.

I fixed this by transforming the query into:

```csharp
repository.Select(_s1 => (_s1 is ConcreteType1) ? (BaseType)_s1
                : (_s1 is ConcreteType2) ? (ConcreteType2)new ConcreteType2 { Foo = ((ConcreteType2)_s1).Foo }
                : default(BaseType));
```

This approach selects the entire record, so it indeed overfetches. Anyways, it seems to be the simplest solution that works in all cases.

I considered selecting just one property instead, but that seems to be hacky and tricky since we cant know for sure whether a property is actually part of the table or just a calculated helper property. Because we’re only selecting the values of the base type and also do not eager load navigation properties, I believe the performance penalty from this overfetching is negligible compared to correctness. Also this seems to be a rare edge case.

However, if another approach is preferred - like the one with just selecting one property-, please let me know and I can reevaluate this.

### DIFF
--- a/src/HotChocolate/Data/src/Data/Projections/Expressions/QueryableProjectionScopeExtensions.cs
+++ b/src/HotChocolate/Data/src/Data/Projections/Expressions/QueryableProjectionScopeExtensions.cs
@@ -35,12 +35,25 @@ public static class QueryableProjectionScopeExtensions
 
             foreach (var val in scope.GetAbstractTypes())
             {
-                var ctor = Expression.New(val.Key);
-                Expression memberInit = Expression.MemberInit(ctor, val.Value);
+                var instance = scope.Instance.Peek();
+                Expression baseExpression;
+                if (val.Value.Count == 0)
+                {
+                    // No field has been selected for this type
+                    // -> Since we need to select something, select all(!) fields
+                    baseExpression = instance;
+                }
+                else
+                {
+                    // One or more fields have been selected for this type
+                    // -> Select only the requested fields.
+                    var ctor = Expression.New(val.Key);
+                    baseExpression = Expression.MemberInit(ctor, val.Value);
+                }
 
                 lastValue = Expression.Condition(
-                    Expression.TypeIs(scope.Instance.Peek(), val.Key),
-                    Expression.Convert(memberInit, scope.RuntimeType),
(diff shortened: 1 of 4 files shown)


michaelstaib:
@N-Olbert we at the moment do not touch anything regarding HotChocolate.Data or GreenDonut.Data as we plan to bring them together. We will pull out the projections engine and rearchitecture it. I can involve you into that process, but it will be somewhere in November as we are focused on Fusion at this point.

N-Olbert:
@michaelstaib sure, ping me then.

Do you think it makes sense to add the (failing or skipped) tests now, so that this case won't be forgotten then?

PR #8740, opened 2025-09-25
[Fusion] Handle invalid @is usages more gracefully

tobias-tengler:
Previously it would just throw a "Stack empty" exception, which didn't point you to the issue.


src/HotChocolate/Fusion-vnext/src/Fusion.Composition/PostMergeValidationRules/IsInvalidFieldRule.cs
```diff
@@ -42,12 +42,8 @@ public void Handle(SchemaEvent @event, CompositionContext context)
...
             var fieldSelectionMapParser = new FieldSelectionMapParser(fieldArgumentValue);
             var fieldSelectionMap = fieldSelectionMapParser.Parse();
-            var inputTypeNode = sourceArgument.Type.ToTypeNode();
-            var inputTypeDefinition = schema.Types[sourceArgument.Type.AsTypeDefinition().Name];
-            var inputType = inputTypeNode.RewriteToType(inputTypeDefinition);
-            var outputTypeNode = sourceField.Type.ToTypeNode();
-            var outputTypeDefinition = schema.Types[sourceField.Type.AsTypeDefinition().Name];
-            var outputType = outputTypeNode.RewriteToType(outputTypeDefinition);
+            var inputType = sourceArgument.Type;
+            var outputType = sourceField.Type;
```
glen-84: Hmm. I don't think that this will work if the field is in another source schema, I think that it needs to reference the type in the composite schema. `RequireInvalidFieldsRule` (similar) has tests for these cross-schema references, but there are no such tests for this rule.

src/HotChocolate/Fusion-vnext/test/Fusion.Composition.Tests/PostMergeValidationRules/IsInvalidFieldRuleTests.cs
```diff
@@ -101,6 +101,86 @@ type Person {
...
+                        name: String
+                    }
+                    """
+                ],
+                [
+                    "The @is directive on argument 'Query.personByIds(ids:)' in schema 'A' "
+                    + "specifies an invalid field selection against the composed schema."
+                ]
+            },
+            // List output type, List input type (valid for Fusion v1 batch lookups).
```
glen-84, line 186: Will this be supported in the executor?

glen-84 (changes requested):

src/HotChocolate/Fusion-vnext/test/Fusion.Composition.Tests/PostMergeValidationRules/IsInvalidFieldRuleTests.cs
tobias-tengler: I talked about it with Michael at some point to make the transition easier, but it will be a big change, so I'm considering it invalid at the moment. Most people in Fusion v1 weren't using the attributes so when auto-converting from v1 to v2 we can hopefully just ignore these batch fields.

src/HotChocolate/Fusion-vnext/src/Fusion.Composition/PostMergeValidationRules/IsInvalidFieldRule.cs
tobias-tengler: Ah makes sense. I'm going to port one of those tests and revert my change!

glen-84 (approved):

PR #8745, opened 2025-09-25
Hotfix: Use ClaimsPrincipal global prefix when generating resolver.

MartynasZilinskas:

- Added missing `global::` prefix for ClaimsPrincipal type.


michaelstaib (approved):

michaelstaib:
Thanks for contributing to HotChocolate!

PR #8747, opened 2025-09-28
Fix tests on Windows

N-Olbert:

- 16 tests currently fail on Windows
- 1 one non-Englisch systems (Input_Should_MatchSnapshotAndType_When_Passed, due to localized formatting)
- this is fixed with this PR

Failing tests:
<details>
  HotChocolate
   AspNetCore
    test
     HotChocolate.AspNetCore.Tests
      HttpGetSchemaMiddlewareTests
       Download_GraphQL_Schema
        Download_GraphQL_Schema(path: "/graphql/schema")
        Download_GraphQL_Schema(path: "/graphql/schema.graphql")
        Download_GraphQL_Schema(path: "/graphql/schema/")
        Download_GraphQL_Schema(path: "/graphql?sdl")
       Download_GraphQL_Schema_Slicing_Args_Enabled
        Download_GraphQL_Schema_Slicing_Args_Enabled(path: "/graphql/schema")
        Download_GraphQL_Schema_Slicing_Args_Enabled(path: "/graphql/schema.graphql")
        Download_GraphQL_Schema_Slicing_Args_Enabled(path: "/graphql/schema/")
        Download_GraphQL_Schema_Slicing_Args_Enabled(path: "/graphql?sdl")
      HttpPostMiddlewareTests
       SingleRequest_Defer_Results
   Fusion-vnext
    test
     HotChocolate.Fusion.Execution.Tests
      Execution
       Serialization
        JsonOperationPlanSerializationTests
         Parse_Plan
         Parse_Plan_With_Node
   MongoDb
    test
     HotChocolate.Types.MongoDb.Tests
      BsonTypeTests
       Input_Should_MatchSnapshotAndType_When_Passed
        Input_Should_MatchSnapshotAndType_When_Passed(fieldName: "double", value: 43,229999999999997, type: typeof(MongoDB.Bson.BsonDouble))
   Primitives
    test
     HotChocolate.Primitives.Tests
      NameUtilsTests
       InvalidName
        InvalidName(name: "$Bar")
        InvalidName(name: "1_Bar")
        InvalidName(name: "1Bar")
        InvalidName(name: "B+ar")
        InvalidName(name: "B/ar")
</details>

### DIFF
--- a/src/HotChocolate/Core/src/Execution/Configuration/RequestExecutorSetup.cs
+++ b/src/HotChocolate/Core/src/Execution/Configuration/RequestExecutorSetup.cs
@@ -34,6 +34,11 @@ public sealed class RequestExecutorSetup
     public RequestExecutorOptions? RequestExecutorOptions { get; set; }
 
     /// <summary>
+    /// Gets or sets the time that the executor manager waits to dispose the schema services.
+    /// </summary>
+    public TimeSpan EvictionTimeout { get; set; } = TimeSpan.FromSeconds(30);
+
+    /// <summary>
     /// Gets the request executor options actions.
     /// This hook is invoke first in the schema creation process.
     /// </summary>
@@ -117,6 +122,7 @@ public sealed class RequestExecutorSetup
         options._onBuildDocumentValidatorHooks.AddRange(_onBuildDocumentValidatorHooks);
         options._pipelineModifiers.AddRange(_pipelineModifiers);
         options._typeModules.AddRange(_typeModules);
+        options.EvictionTimeout = EvictionTimeout;
 
         if (DefaultPipelineFactory is not null)
         {
--- a/src/HotChocolate/Core/src/Execution/HotChocolate.Execution.csproj
+++ b/src/HotChocolate/Core/src/Execution/HotChocolate.Execution.csproj
@@ -24,6 +24,7 @@
     <InternalsVisibleTo Include="HotChocolate.Execution.Projections" />
     <InternalsVisibleTo Include="HotChocolate.Execution.Tests" />
     <InternalsVisibleTo Include="HotChocolate.Fusion.Tests" />
+    <InternalsVisibleTo Include="HotChocolate.Types.Tests" />
     <InternalsVisibleTo Include="HotChocolate.Fusion" />
     <InternalsVisibleTo Include="HotChocolate.Types.Mutations" />
     <InternalsVisibleTo Include="StrawberryShake.CodeGeneration" />
@@ -218,6 +219,15 @@
     </Compile>
     <Compile Update="DependencyInjection\RequestExecutorBuilderExtensions.Services.cs">
       <DependentUpon>RequestExecutorBuilderExtensions.cs</DependentUpon>
+    </Compile>
+    <Compile Update="RequestExecutorManager.Events.cs">
+      <DependentUpon>RequestExecutorManager.cs</DependentUpon>
+    </Compile>
+    <Compile Update="RequestExecutorManager.Hooks.cs">
+      <DependentUpon>RequestExecutorManager.cs</DependentUpon>
+    </Compile>
+    <Compile Update="RequestExecutorManager.Warmup.cs">
+      <DependentUpon>RequestExecutorManager.cs</DependentUpon>
     </Compile>
   </ItemGroup>
 
--- a/src/HotChocolate/Core/src/Execution/RequestExecutorManager.cs
+++ b/src/HotChocolate/Core/src/Execution/RequestExecutorManager.cs
@@ -27,7 +27,7 @@ internal sealed partial class RequestExecutorManager
     : IRequestExecutorManager
     , IRequestExecutorEvents
     , IRequestExecutorWarmup
-    , IDisposable
+    , IAsyncDisposable
 {
     private readonly CancellationTokenSource _cts = new();
     private readonly ConcurrentDictionary<string, SemaphoreSlim> _semaphoreBySchema = new();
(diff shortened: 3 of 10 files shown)


src/HotChocolate/AspNetCore/test/AspNetCore.Tests/HttpGetSchemaMiddlewareTests.cs
```diff
@@ -54,6 +54,8 @@ public async Task Download_GraphQL_Schema(string path)
 
         // assert
         Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        response.Headers.Remove("ETag");
```
N-Olbert, line 57: ETag and ContentLength differ based on whether the response contains \n or \r\n as new line delimiter. I removed them as I think these values are out-of-scope for the actual test case.

N-Olbert:
@glen-84 I use LF with no autocrlf. Using CRLF with autocrlf opens a different can of worms (I tried it out) as that implicitly changes all multi-line raw string literals. This will change f. e. the operation hash of such operations, causing a few (around 10 or so) other tests to fail.

Anyhow, I do not see a single change here that shouldn't be required for CRLF-environments. If any, such environments will require additional adjustments.

glen-84:
> This will change f. e. the operation hash of such operations, causing a few (around 10 or so) other tests to fail.

Ya, this is a pain, but if we decided that operation hashes _should_ vary depending on platform (not sure about this), then one option would be to scrub hashes from the snapshots.

> Anyhow, I do not see a single change here that shouldn't be required for CRLF-environments. If any, such environments will require additional adjustments.

We wouldn't need the special handling in the CookieCrumble Snapshot class, or any of the `ReplaceLineEndings` calls.

N-Olbert:
> but if we decided that operation hashes should vary depending on platform (not sure about this)

Probably it would indeed be beneficial to have a linefeed-agnostic hash: If I recall correctly, the `OperationHash` is basically  
the key of the `OperationCacheMiddleware`, so currently the same request once with \n and once with \r\n will take two cache slots. However, I think getting this would tricky and probably also too computationally  intensive, since I think user data / value node literals may not be modified at all.  
So (assume that all line feeds are not escaped, I just added them for better understanding):

```graphql
foo(arg: "abc") { \n
    p1 { \n
        bar \n
    } \n
} \n
```
is indeed identical to:

```graphql
foo(arg: "abc") { \r\n
    p1 { \r\n
        bar \r\n
    } \r\n
} \r\n
```
but

```graphql
foo(arg: "a\nb\nc\n") { \n
    p1 { \n
        bar \n
    } \n
} \n
```
is not identical to:

```graphql
foo(arg: "a\r\nb\r\nc\r\n") { \r \n
    p1 { \r\n
        bar \r\n
    } \r\n
} \r\n
```

Maybe its worth experimenting with it...

> We wouldn't need the special handling in the CookieCrumble Snapshot class, or any of the ReplaceLineEndings calls. 

A lot of snaps contain escaped new line delimiters, currently only `\n` (except for the http-multipart-snaps, those contain \r\n as this delimiter is defined by the spec), so I think the change in CookieCrumble would still be necessary.

glen-84:
> However, I think getting this would tricky and probably also too computationally intensive, since I think user data / value node literals may not be modified at all.

Why would they need to be modified? Wouldn't we just replace literal CRLFs with LFs in the byte array before hashing?

> A lot of snaps contain escaped new line delimiters, currently only \n (except for the http-multipart-snaps, those contain \r\n as this delimiter is defined by the spec), so I think the change in CookieCrumble would still be necessary.

Some of these would be caused by `NormalizeLineBreaks()`, others because the resource (graphql file) in Git would currently be checked out with LF. I'm not sure how many other cases there are though.

Certainly not a trivial task. 🙂

glen-84:
@N-Olbert Thanks!

N-Olbert:
> Why would they need to be modified? Wouldn't we just replace literal CRLFs with LFs in the byte array before hashing?

I dont know...

If it were just a plain hash, replacing CRLFs with LFs in the byte stream would work, but since we’re using an md5/sha hash of the document / request as a unique identifier (without an explicit equality check) in a lot of cache layers, I think that approach is too simple.

F. e. we use the operation / document hash as cache key within the cache middlewares:
https://github.com/ChilliCream/graphql-platform/blob/906d9d3f2fba45b581ea56ff7533f208e80c8e6e/src/HotChocolate/Core/src/Execution.Pipeline/DocumentCacheMiddleware.cs#L65-L66
https://github.com/ChilliCream/graphql-platform/blob/906d9d3f2fba45b581ea56ff7533f208e80c8e6e/src/HotChocolate/Fusion-vnext/src/Fusion.Execution/Execution/Pipeline/OperationPlanCacheMiddleware.cs#L30-L37
https://github.com/ChilliCream/graphql-platform/blob/906d9d3f2fba45b581ea56ff7533f208e80c8e6e/src/HotChocolate/Core/src/Execution/Pipeline/OperationCacheMiddleware.cs#L38-L44

I think the core keypoints are:
-  For structural newlines (those between tokens in the GraphQL document), normalizing \r\n → \n before hashing makes sense and won`t change anything during execution, so
```
foo { 
  bar 
}

```
could be treated as the same with either \n or \r\n or any permutation of those after line 1, 2, 3.

- But for string literals, the line endings are part of the actual request data, so they must remain unchanged, since `foo(arg: "a\nb\nc\n")` and `foo(arg: "a\r\nb\r\nc\r\n")` aren`t interchangable, as the resolver logic may depend on the exact line feed characters (e.g. "\r\n" repeated 3 times means X, while "\n" repeated 3 times means Y)

So while I see the appeal of the normalization, I don’t think it can be done reliably at the byte level without risking that distinct operations collapse into the same hash - but maybe I oversee something.

glen-84:
There are no literal CRLFs/LFs in a regular GraphQL string, they are escaped. However, block strings would be an issue.

Maybe we need a specialized syntax printer that builds a string without whitespace between tokens, and only preserves whitespace inside strings.

PR #8758, opened 2025-10-02
Make request executor creation eager by default

tobias-tengler:


src/HotChocolate/Core/src/Execution.Abstractions/CombinedServiceProvider.cs
```diff
@@ -85,14 +85,19 @@ public CombinedServiceProvider(IServiceProvider first, IServiceProvider second)
                 }
             }
 
+            // This should be fine, as we ensure the constructor is included,
+            // when registering services with the service collection.
+#pragma warning disable IL3050
             var array = Array.CreateInstance(elementType, buffer.Count);
+#pragma warning restore IL3050
```
tobias-tengler: Sketchy

michaelstaib: This whole class is on my kill list for 16. I think the convenience here was a mistake.

michaelstaib: Let's put a task on the backlog for this

PR #8759, opened 2025-10-02
Fix: Enum-deserialization from _Any-type with ReferenceResolver

N-Olbert:

- Enums can appear as `StringValue` in `_Any`, this was not considered in the `ArgumentParser`
- Detail 2


StefanGies99:
Hi @N-Olbert any idea when this is going to be released?

N-Olbert:
Hi @StefanGies99, unfortunately I cant help with that. This PR hasn’t been triaged for review yet, so I’m not even sure if it will make it into main (though I suspect it will).
Probably @michaelstaib has more insight on this and maybe can answer your question.

PR #8764, opened 2025-10-05
[Fusion] Update introspection directive to use 'args' and add corresp…

Hanskrogh:
**Introspection logic update:**

* Changed the resolver in `__Directive.cs` to use `"args"` instead of `"arguments"` when setting directive arguments, ensuring compatibility with the GraphQL introspection query format.

**Testing and verification:**

* Added a new test `Directives_Should_Include_Args` in `IntrospectionTests.cs` to confirm that the introspection query returns directive arguments as expected.
* Introduced a snapshot file `IntrospectionTests.Directives_Should_Include_Args_SkipIncludeDirectiveArgs.yaml` to capture and validate the expected response for the new test case.

### DIFF
--- a/src/HotChocolate/Fusion-vnext/src/Fusion.Execution/Execution/Introspection/__Directive.cs
+++ b/src/HotChocolate/Fusion-vnext/src/Fusion.Execution/Execution/Introspection/__Directive.cs
@@ -27,7 +27,7 @@ internal sealed class __Directive : ITypeResolverInterceptor
             case "locations":
                 features.Set(new ResolveFieldValue(Locations));
                 break;
-            case "arguments":
+            case "args":
                 features.Set(new ResolveFieldValue(Arguments));
                 break;
         }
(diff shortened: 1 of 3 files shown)


tobias-tengler:
Hey @Hanskrogh :)
Good catch!
The snapshot of `IntrospectionTests.IntrospectionQueries_IntrospectionQuery` should also have changed, since we're querying for the `args` there, but missed that they're currently returning `null`.
Since this test already covers it, I think you can also remove your newly added test again.

Hanskrogh:
> Hey @Hanskrogh :) Good catch! The snapshot of `IntrospectionTests.IntrospectionQueries_IntrospectionQuery` should also have changed, since we're querying for the `args` there, but missed that they're currently returning `null`. Since this test already covers it, I think you can also remove your newly added test again.

@tobias-tengler Check out the new commit :) Also verified, that all tests are working now! ;)

tobias-tengler (approved):
@Hanskrogh Looks good! Thank you for your contribution :)
We'll probably release this tomorrow with a new v16 preview.

PR #8765, opened 2025-10-05
[Fusion] Add support for directives in schema composition fusion-v2

Hanskrogh:
Add support for merging directives in Fusion-v2 schemas

Took a first stab at handling directives in the SchemaMerger. Not sure if this is the final behavior (might not even be fully specified yet), but thought it could serve as a reference point.  

@tobias-tengler  feel free to close if this doesn’t fit or if I’ve missed something. This is my first dive into the SchemaMerger, so I might have overlooked a few bits.


glen-84:
Hi @Hanskrogh,

It's always best to open an issue or discussion before submitting a PR. There are reasons why this is not yet implemented – firstly, it's unspecified, and secondly, we are discussing a configuration-based approach internally. We won't be merging every directive.

PR #8766, opened 2025-10-05
fix: Add support list value type in requirements

Hanskrogh:
Right now, when a field has a list type as a `@require`, the gateway just resolves to null. This PR adds proper support so lists behave as expected instead of silently dropping to null.

It’s still a first pass, but should fix the most obvious case. Feedback welcome!

### DIFF
--- a/src/HotChocolate/Fusion-vnext/src/Fusion.Execution/Execution/Results/SelectionMapExecutor.cs
+++ b/src/HotChocolate/Fusion-vnext/src/Fusion.Execution/Execution/Results/SelectionMapExecutor.cs
@@ -79,9 +79,29 @@ internal static class ResultDataMapper
                 return NullValueNode.Default;
             }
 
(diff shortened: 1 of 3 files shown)


michaelstaib:
We are at the moment working on a big change to the underlying system so that we cannot merge this one just yet. Once https://github.com/ChilliCream/graphql-platform/pull/8668 is completed we can have a look at this one.

Hanskrogh:
Should i update the PR with the new changes or is it on your internal task list to implement support for list requirements in v2?

michaelstaib:
@Hanskrogh yes, update it on top of the new result structure and we merge it.

Hanskrogh:
@michaelstaib Apologies for the mess here. My local branch ended up in a pretty broken state, which caused the problems you saw. I’ve cleaned things up and it should be sorted now.

michaelstaib:
dont worry, we squash anyway at the end.

PR #8771, opened 2025-10-06
Throw when trying to resolve unknown schema

tobias-tengler:
I just spend 30min looking for the misconfiguration behind the following error:

> The schema builder was unable to identify the query type of the schema.

Turns out the test I was working on was requesting a different schema name from the one that was registered...

This PR now throws an exception when trying to request a schema that does not exist, to avoid confusion.

### DIFF
--- a/src/HotChocolate/Core/test/Execution.Tests/RequestExecutorManagerTests.cs
+++ b/src/HotChocolate/Core/test/Execution.Tests/RequestExecutorManagerTests.cs
@@ -4,8 +4,28 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace HotChocolate.Execution;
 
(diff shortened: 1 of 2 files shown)


src/HotChocolate/Core/src/Execution/RequestExecutorManager.cs
```diff
@@ -92,6 +92,11 @@ public async ValueTask<IRequestExecutor> GetExecutorAsync(
                 return re.Executor;
             }
 
+            if (!SchemaNames.Contains(schemaName))
```
tobias-tengler, line 95: `SchemaNames` is just an ImmutableArray and not a HashSet, but this check happens behind a semaphore, so not on the hot path, and the number of items in the array is expected to be low.

PR #8773, opened 2025-10-06
[Fusion] Ignore inapplicable type refinements on abstract types

Coelho04:
Reported here https://github.com/ChilliCream/graphql-platform/issues/6727#issuecomment-3367589276

An interface can be distributed across multiple source schemas, and each schema may implement that interface with different concrete types. The planner currently fails when you specify a type refinement for a concrete type that doesn't exist in the source schema that provides the field returning the distributed interface.
Since a source schema can only return the concrete types it actually defines, we can safely ignore inapplicable type refinements during planning.


tobias-tengler (approved):
Looks good! Thank you for your contribution :)
I've took the liberty to clean up a bit and fix the `Interface_Field_With_Only_Type_Refinements_Exclusive_To_Other_Schema` test.

If you sign the CLA and the tests have passed, I can merge this.

PR #8797, opened 2025-10-12
Update RabbitMQ.Client to version 7.0.0

### LINKED ISSUE #8381: Subscriptions.RabbitMQ Dependency Upgrade
### Product

Hot Chocolate

The `HotChocolate.Subscriptions.RabbitMQ` Package depends on the [RabbitMQ.Client Package Version 6.4.0](https://www.nuget.org/packages/RabbitMQ.Client/6.4.0) which is 3 years old as of now.

When trying to combine RabbitMQ Subscription Management of HotChocolate on the Same Server that implements an Event-Driven Architecture using [MassTransit](https://www.nuget.org/packages/MassTransit.RabbitMQ/8.5.0) for example, also using RabbitMQ as its message broker you will run into Version-Conflicts because of their dependent Package.

But the kicker is, you won't even realize this since the `HotChocolate.Subscriptions.RabbitMQ` Package, from my understanding, uses reflection to call Functions on the `IModel` Interface of the `RabbitMq.Client` Package, which no longer exists in its current versions.

### The solution you'd like

The dependency to `RabbitMQ.Client` upgraded to it's current version (7.1.2) and maybe try to use it without reflection so you would get Compile-Time Errors when working in combination with other Versions of RabbitMQ.Client from other packages.

ittennull:

- Update RabbitMQ.Client to version 7.0.0
- Refactor topic implementations to always use IAsyncDisposable

Closes #8381 

Old version of RabbitMQ.Client creates runtime errors if a projects uses 7.* version of RabbitMQ.Client or any dependency that rely on it, for example Masstransit


tobias-tengler:
Thank you for the contribution @ittennull :)

Some tests are still failing. It would be great if you could get them to pass.

ittennull:
I fixed one test but all the RabbitMQ tests are red because the test library Squadron.RabbitMQ also uses outdated RMQ client which is not compatible with the new version. They don't have an update for it yet.
I ran the tests with RMQ running on localhost (I replaced connection values in test) and everything was green. However, I understand that it's not sufficient. 
Maybe we can look into NOT using Squadron for RMQ?

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQTopic.cs
```diff
@@ -11,32 +11,54 @@ internal sealed class RabbitMQTopic<TMessage> : DefaultTopic<TMessage>
...
 
-    protected override async ValueTask<IDisposable> OnConnectAsync(
+    protected override async ValueTask<IAsyncDisposable> OnConnectAsync(
         CancellationToken cancellationToken)
     {
         // We ensure that the processing is not started before the context is fully initialized.
-        Debug.Assert(_connection != null, "_connection != null");
-        Debug.Assert(_connection != null, "_serializer != null");
+        Debug.Assert(_connection != null);
+        Debug.Assert(_serializer != null);
```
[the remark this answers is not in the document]
ittennull: That's exactly what I do

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/IRabbitMQConnection.cs
glen-84, line 1: I've corrected the spelling of the file name on `main` (in #8798). Please update it when merging `main`.

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQTopic.cs
```diff
@@ -45,64 +67,47 @@ Task Received(object sender, BasicDeliverEventArgs args)
...
+            cancellationToken);
 
         DiagnosticEvents.ProviderTopicInfo(Name, RabbitMQTopic_ConnectAsync_SubscribedToRabbitMQ);
 
-        return new Subscription(() =>
+        return new Subscription(async () =>
         {
-            channel.BasicCancelNoWait(consumerTag);
-            consumer.Received -= Received;
+            await channel.BasicCancelAsync(consumer.ConsumerTags.First(), noWait: false, cancellationToken);
```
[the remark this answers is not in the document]
ittennull: At this point we already connected to a queue, hence there is a consumer tag

tobias-tengler:
Thanks for fixing up the tests!
I like the switch to Testcontainers, but I'll check with the team first to confirm there wasn't a specific reason we chose Squadron originally.

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQTopic.cs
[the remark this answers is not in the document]
tobias-tengler: Copilot must've sipped on the silly juice...

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/IRabbitMQConnection.cs
ittennull: I merged `main` branch into my branch

ittennull:
>I'll check with the team first to confirm there wasn't a specific reason we chose Squadron originally.

TestContainers work better here because the package only starts and stops containers and it doesn't depend on RabbitMq.Client library at all, so it's safer for any future updates

ittennull:
>  I'll check with the team first to confirm there wasn't a specific reason we chose Squadron originally.

@tobias-tengler Hi, did you have a chance to confer with the team yet?

tobias-tengler:
@ittennull Yes, we're going to stick with Squadron for now. 
We just released a new version that should be compatible: https://github.com/SwissLife-OSS/squadron/releases/tag/0.26.0
Could you please undo your TestContainer changes and update to that Squadron version instead?

Also, please downgrade RabbitMQ.Client to just 7.0.0. We do not want to force people to use a specific minor/patch version.

ittennull:
OK, will do. The update of Squadron [failed](https://github.com/SwissLife-OSS/squadron/actions/runs/18465140991) though

ittennull:
I saw that you updated Squadron to support RMQ 7 but the build fails because of the build scripts. I created a PR to fix it but nobody reviews it. I even wrote an email to the person who created these build scripts - no response.
I'm not sure if Squadron is a good dependency, in 6 years there are 100 commits, no updates, the github insights shows that the project is practically dead, the authors don't review PRs. Perhaps for the sake of hotchocolate you guys can re-consider if moving to another testing library is beter

tobias-tengler:
@ittennull Good catch with the .slnx. I'll just undo my change to the solution file in the Squadron repository. Let's hope this fixes the issue. 
In the short term @michaelstaib wants to move Squadron to this repository, as it was primarily built as test tooling for Hot Chocolate. This would also get rid of dependency issues like this.
Sorry for the delay here!

tobias-tengler:
@ittennull Squadron v26.0.0 is now available :)

ittennull:
@tobias-tengler I updated the package. Could you approve the workflow and check the PR again?

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/DependencyInjection/RabbitMQPubSubExtensions.cs
tobias-tengler: The `rabbitMqSubscriptionOptions` probably need to be forwarded here:

```suggestion
            .AddRabbitMQSubscriptionPublisher(connectionFactory, options, rabbitMqSubscriptionOptions)
```

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQPubSub.cs
```diff
@@ -1,25 +1,31 @@
...
 internal sealed class RabbitMQPubSub : DefaultPubSub
 {
+    private readonly SemaphoreSlim _semaphore = new(1, 1);
     private readonly IRabbitMQConnection _connection;
     private readonly IMessageSerializer _serializer;
+    private readonly RabbitMQSubscriptionOptions _rabbitMqSubscriptionOptions;
     private readonly string _completed;
     private readonly int _topicBufferCapacity;
     private readonly TopicBufferFullMode _topicBufferFullMode;
+    private readonly BasicProperties _publishBasicProperties = new() { Persistent = true };
```
tobias-tengler, line 16: The default seems to be transient. What's your reasoning for this change? Is this something a user should be able to configure?

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQPubSub.cs
```diff
@@ -48,19 +54,38 @@ protected override DefaultTopic<TMessage> OnCreateTopic<TMessage>(
...
         var body = Encoding.UTF8.GetBytes(message);
+        var channel = await _connection.GetChannelAsync(cancellationToken).ConfigureAwait(false);
 
-        channel.BasicPublish(
-            exchange: formattedTopic,
-            routingKey: string.Empty,
-            mandatory: false,
-            properties,
-            body);
+        await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
tobias-tengler, line 65: Why do we have this synchronization now?

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/DependencyInjection/RabbitMQPubSubExtensions.cs
ittennull: good catch! Thanks. Fixed it

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQPubSub.cs
ittennull: The queue that receives messages that are published with this BasicProperties is marked as durable. It was like this even before my PR. So I thought that that was the intention - to save messages so that if a RMQ node restarts, the messages are available. That makes sense to me, seems like a good default, although maybe somebody else wouldn't care about it and non-durable queues would do for their case too. Anyway, let's assume we want persistence. 
According to the documentation:
- _In environments and use cases where durability is important, applications must use durable queues and make sure that publishers mark published messages as persisted._ [durability](https://www.rabbitmq.com/docs/queues#durability)
- _When RabbitMQ quits or crashes it will forget the queues and messages unless you tell it not to. Two things are required to make sure that messages aren't lost: we need to mark both the queue and messages as durable._ [message-durability](https://www.rabbitmq.com/tutorials/tutorial-two-dotnet#message-durability)

That why I changed it.

Should we make it configurable? Not sure. How much configurable? We need to change both values: for the queue and the message. But what if somebody has multiple graphql subscriptions, one of them is critical and no messages must be lost, others - nobody cares about them. Then we need to allow configuration per each graphql subscription. And what about other parameters. Can be too much configuration

What do you think? We can leave it like this until somebody comes in with a use-case: we want to adjust this because of that

src/HotChocolate/Core/src/Subscriptions.RabbitMQ/RabbitMQPubSub.cs
ittennull: The code follows [recommendations](https://www.rabbitmq.com/client-libraries/dotnet-api-guide#connection-and-channel-lifespan): connections and channels are long-lived and shared. When an app has multiple graphql subscriptions, it can publish multiple messages at the same time. Using a channel by many threads without a synchronization will lead to errors. So I follow the [guide](https://www.rabbitmq.com/client-libraries/dotnet-api-guide#concurrency-channel-sharing) and synchronize the access to the channel

ittennull:
I also marked the queue as auto-delete because I noticed that we have thousands of messages and queues lying around with zero consumers. That happens when a graphql subscription finishes and a consumer unsubscribes. A queue is exclusive and isn't deleted because the connection that created it still lives and serves other new queues. So the only way to clean RMQ memory and storage is to restart your app

PR #8845, opened 2025-10-23
Add support for shareable to the source generator.

michaelstaib:


src/HotChocolate/Core/test/Types.Tests/Types/Composite/ShareableTests.cs
glen-84, line 1: `Shareable`

src/HotChocolate/Core/src/Types.Analyzers/Errors.cs
```diff
@@ -112,4 +112,130 @@ public static class Errors
...
+            messageFormat: "The [Shareable] attribute on properties and methods must not specify the 'scoped' argument",
+            category: "TypeSystem",
+            DiagnosticSeverity.Error,
+            isEnabledByDefault: true);
+
+    public static readonly DiagnosticDescriptor NodeResolverIdParameter =
+        new(
+            id: "HC0104",
+            title: "NodeResolver First Parameter Must Be Named 'id'",
+            messageFormat: "The first parameter of a node resolver must be the node id and must be called 'id'",
```
glen-84:
```suggestion
            messageFormat: "The first parameter of a node resolver must be the node ID and must be named 'id'",
```

PR #8870, opened 2025-11-05
[Fusion] Fix findings

tobias-tengler:
- Update Variable Coercion to handle missing required properties and `@oneOf`
- Initialize ScalarSerializationType for built-in scalars
- Fix error code not being properly parsed from source schema result

### DIFF
--- a/src/HotChocolate/Core/src/Types.Abstractions/Types/Extensions/HotChocolateTypesAbstractionsScalarDefinitionExtensions.cs
+++ b/src/HotChocolate/Core/src/Types.Abstractions/Types/Extensions/HotChocolateTypesAbstractionsScalarDefinitionExtensions.cs
@@ -1,11 +1,25 @@
+using System.Collections.Frozen;
+
 namespace HotChocolate.Types;
 
 public static class HotChocolateTypesAbstractionsScalarDefinitionExtensions
 {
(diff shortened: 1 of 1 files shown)


src/HotChocolate/Mutable/src/Types.Mutable/BuiltIns/BuiltIns.cs
```diff
@@ -140,7 +160,6 @@ public static bool IsBuiltInDirective(string name)
             Skip.Name => true,
             Deprecated.Name => true,
             SpecifiedBy.Name => true,
-            OneOf.Name => true,
```
tobias-tengler: TODO: Should `@oneOf` still be serialized into the schema or does it behave like the other spec directives now?

src/HotChocolate/Fusion-vnext/src/Fusion.Composition/SourceSchemaMerger.cs
```diff
@@ -233,6 +241,20 @@ private void AddFusionLookupDirectives(MutableSchemaDefinition mergedSchema)
         }
     }
 
+    private void AddOneOfDirective(
```
tobias-tengler: TODO: Discuss with Michael

PR #8874, opened 2025-11-06
[Fusion] Added support for merging `@serializeAs` directives

glen-84:

- [Fusion] Added support for merging `@serializeAs` directives.

### DIFF
--- a/src/HotChocolate/Fusion-vnext/src/Fusion.Composition/DirectiveMergers/OneOfDirectiveMerger.cs
+++ b/src/HotChocolate/Fusion-vnext/src/Fusion.Composition/DirectiveMergers/OneOfDirectiveMerger.cs
@@ -10,10 +10,7 @@ namespace HotChocolate.Fusion.DirectiveMergers;
 internal class OneOfDirectiveMerger(DirectiveMergeBehavior mergeBehavior)
     : DirectiveMergerBase(mergeBehavior)
 {
-    public override string DirectiveName { get; } =
-        mergeBehavior is DirectiveMergeBehavior.IncludePrivate
-            ? $"fusion__{DirectiveNames.OneOf}"
-            : DirectiveNames.OneOf;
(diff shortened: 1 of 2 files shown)


src/HotChocolate/Fusion-vnext/src/Fusion.Composition/DirectiveMergers/OneOfDirectiveMerger.cs
```diff
@@ -0,0 +1,51 @@
...
+using DirectiveNames = HotChocolate.Fusion.WellKnownDirectiveNames;
+
+namespace HotChocolate.Fusion.DirectiveMergers;
+
+internal class OneOfDirectiveMerger(DirectiveMergeBehavior mergeBehavior)
+    : DirectiveMergerBase(mergeBehavior)
+{
+    public override string DirectiveName { get; } =
+        mergeBehavior is DirectiveMergeBehavior.IncludePrivate
+            ? $"fusion__{DirectiveNames.OneOf}"
```
tobias-tengler: I think this makes no sense for `@oneOf` right?

glen-84: Ya, kinda, I left it like this for `serializeAs` as well, even though both of them have their behaviour fixed and not configurable ...

https://github.com/ChilliCream/graphql-platform/pull/8874/files#diff-02a127093f9b97262c0c82dc16eefda4b74893062d018f2850909902756ab7efR54-R55

Do you think that I should change it?

```cs
public override string DirectiveName { get; } = DirectiveNames.OneOf;
```

glen-84: I guess I'll just remove it, and the unnecessary `MergeDirectiveDefinition` override for `serializeAs`.

PR #8889, opened 2025-11-11
Migration docs for PR 8504

N-Olbert:

- Documented breaking change of #8504 
- Detail 2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
         name: Setup node
         with:
           node-version: 22
-      - run: npm install -g markdownlint-cli2
+      - run: npm install -g markdownlint-cli2@0.19.0
         name: Install markdownlint-cli2
       - run: markdownlint-cli2 "*.md" "website/src/**/*.md"
         name: run Markdownlint
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -28,3 +28,5 @@ MD033:
     - Video
 MD036: false # Rationale: Possibly too strict, used frequently in documentation.
 MD059: false # Rationale: Already used in many places in the documentation.
+MD060:
+  style: aligned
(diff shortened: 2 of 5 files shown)


website/src/docs/hotchocolate/v16/migrating/migrate-from-15-to-16.md
```diff
@@ -212,6 +212,21 @@ In addition, the default output for such errors has been standardized: earlier,
 }
 ```
 
+## Generic `ID<Type>`-attribute now infers the actual GraphQL type name
+
+Previously, `[ID<Type>]` used the CLR type name (`nameof(Type)`), even when a different GraphQL type name was configured via `[GraphQLName]` or `descriptor.Name()`.
+It now uses the actual GraphQL type name if one is defined, f. e.:
```
glen-84:
```suggestion
It now uses the actual GraphQL type name if one is defined, for example:
```

Probably better to spell it out here. 🙂

website/src/docs/hotchocolate/v16/migrating/migrate-from-15-to-16.md
```diff
@@ -212,6 +212,21 @@ In addition, the default output for such errors has been standardized: earlier,
...
+
+```csharp
+[GraphQLName("Book")]
+public sealed class BookDTO { }
```
glen-84:
```suggestion
public sealed class BookDTO;
```

glen-84 (changes requested):

website/src/docs/hotchocolate/v16/migrating/migrate-from-15-to-16.md
N-Olbert: I changed it to be a valid GraphQL type by adding two fields.
This seemed like a better option because I explicitly state that every type that is used as a type argument must be a valid GraphQL type.

glen-84 (approved):

glen-84:
I'm going to fix the markdownlint issues on `main` before I merge this.

glen-84:
@N-Olbert Thanks!

PR #8928, opened 2025-11-19
Fix test failures on non-english systems

N-Olbert:

- Force specific culture for localized compile errors in analyzers tests
- The following tests currently fail on non-English systems because the emitted compile errors are localized and therefore differ from the snapshot:
```
HotChocolate.Types.Analyzers.Tests
      BindMemberAnalyzerTests
       BindMember_WithNameof_InvalidMember_RaisesError
      NodeResolverAnalyzerTests
       NodeResolver_NonStatic_WithIdAttribute_RaisesError
      ObjectTypeTests
       GraphQLType_String_On_Parameter_MatchesSnapshot
      QueryContextProjectionAnalyzerTests
       QueryContext_MultipleAttributes_OnlyUseProjectionFlagged
       QueryContext_WithoutUseProjection_NoError
       QueryContext_WithUseProjection_RaisesError
       QueryContext_WithUseProjections_RaisesError
      ResolverTests
       Resolver_Parameter_With_Two_Attribute
```


src/HotChocolate/Core/test/Types.Analyzers.Tests/TestHelper.cs
```diff
@@ -314,4 +318,23 @@ private static void AddDiagnosticsToSnapshot(
 
     [GeneratedRegex("MiddlewareFactories([a-z0-9]{32})")]
     private static partial Regex MiddlewareFactoryHashRegex();
+
+    private sealed class UseSpecificCulture : IDisposable
```
michaelstaib: Lets run the test in .NET 10 and use a module initializer instead ... 

```
internal static class ModuleInitializer
{
    [ModuleInitializer]
    internal static void Initialize()
    {
        CultureInfo.DefaultThreadCurrentCulture = CultureInfo.InvariantCulture;
        CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
    }
}
```

otherwise we could get into thread issues as test run in parallel and CultureInfo.CurrentCulture is static. So better we pin it for the whole test run.

N-Olbert: I am pretty sure that this will not cause threading issues, as it ist basically a (cached) `AsyncLocal` and exactly the same thing ASP.NET core does:

https://github.com/dotnet/aspnetcore/blob/0d1029654a05e1685eea0b7689b42ca07b2af201/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs#L123-L127

and xunit wihtin their samples:

https://github.com/xunit/xunit/blob/master/test/test.utility/CultureAwareTesting/CulturedXunitTestCase.cs#L67-L76

But yeah, using a module initializer is definitely nicer 👍 

Question:
As suggested I added the initializer only to the current test project. Would it make sense in your opinion to extract it into a single file and include this file (via a the Directory.Build.props) in *every* (non-cookiecrumble) test project? Varying culture is probably not something most people have in mind when writing tests, so making it consistent by default might avoid such bugs in the future.

michaelstaib: Could be ... I did not look at the implementation of `CurrentCulture`. But I think still its clearer to have all tests always run in the US culture.

PR #8931, opened 2025-11-20
Adds ResultDocument to HotChocolate Core

michaelstaib:

PR #8972, opened 2025-12-04
Prefer comparing to interfaces

tscott-servicetitan:
### Summary

When using a separate (custom) `IResponseStream` this check will fail. Rather, we can allow this check to pass as long as `IResponseStream` is implemented.

#### Note

Not sure is this is safe/ok. I came across this issue in my own unit tests after implementing my own custom `IResponseStream`. Within the middleware, I will replace the original (context.Result) ResponseStream within a wrapper of my own making.


michaelstaib (changes requested):
Not sure if this is a good idea... there are several guarantees around ResponseStream  and OperationResult. Lets wait on this until we are further with 16 ... for 15 we cannot change it.  For 16 I want to have a clearer picture where we are heading with result types as we are currently working on this: https://github.com/ChilliCream/graphql-platform/pull/8931

tscott-servicetitan:
Makes sense, I thought that may be the case. Thank you for taking a look!

PR #9000, opened 2025-12-14
Source Generator: Infer XML documentation for parameters

N-Olbert:

- Infer XML documentation for parameters within the source generator
- Improved inheritdoc-handling and corresponding tests
- Added tests


src/HotChocolate/Core/src/Types.Analyzers/Helpers/SymbolExtensions.cs
```diff
@@ -239,42 +244,127 @@ public static MethodDescription GetDescription(this IMethodSymbol method, Compil
...
+
                 if (inheritedDoc != null)
                 {
-                    inheritdocElement.ReplaceWith(inheritedDoc);
+                    // Inheritdoc has no well-defined specification, f. e. see:
+                    // https://github.com/dotnet/csharplang/issues/313,
+                    // https://github.com/dotnet/roslyn/issues/68879,
+                    // https://github.com/dotnet/roslyn/discussions/50192 and many others
+
+                    // For this reason we are programming our own lookup version, which is also the recommended way:
```
N-Olbert, line 295: Please note:  As an alternative we could also call the [internal](https://github.com/dotnet/roslyn/discussions/50192#discussioncomment-254793) roslyn implementation with reflection, which is what docfx does, see: 

https://github.com/dotnet/docfx/blob/107e3ef0c0c4db7ac984a3d15e40f7942531f255/src/Docfx.Dotnet/ExtensionMethods/ISymbolExtensions.cs#L96

N-Olbert: No, we can't... this won't work. See https://github.com/ChilliCream/graphql-platform/pull/9041

PR #9056, opened 2026-01-21
Add XML doc inference opt-out control

N-Olbert:

- Introduced GraphQLIgnoreXmlDocumentationAttribute to allow the developer to control XML documentation inference
- Adds analyzer HC0106 to warn if XML documentation is not enabled when inference is active (which is the default)

This may introduce a merge conflict with #9000 for the tests; if so I’ll update whichever branch will be merged last.


michaelstaib:
I have implemented this functionality slightly different here #9299

PR #9107, opened 2026-02-11
[Nitro CLI] Add --archive option to fusion settings set command

tobias-tengler:


src/Nitro/CommandLine/src/CommandLine/Commands/Fusion/FusionSettingsSetCommand.cs
```diff
@@ -90,36 +97,79 @@ private static async Task<int> ExecuteAsync(
                 compositionSettings.Merger.EnableGlobalObjectIdentification = enableGlobalObjectIdentification;
                 break;
 
-            case SettingNames.IncludeSatisfiabilityPaths:
```
tobias-tengler, line 93: @glen-84 for me the `--include-satisfiability-paths` option is very much ephemeral. I don't think this is something that should get embedded (remembered) by the archive. That's why I removed it here.
What was your reasoning for this?

glen-84: I agree. I must not have been thinking clearly.

PR #9324, opened 2026-03-08
[GreenDonut] Properly support ToBatchPageAsync with valueSelector

tobias-tengler:


src/GreenDonut/src/GreenDonut.Data.Primitives/ElementCursorPage.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        var index = Items.IndexOf(item);
+
+        if (index < 0)
+        {
+            throw new ArgumentException("The specified item does not belong to this page.", nameof(item));
+        }
+
+        return _elements[index];
+    }
```
tobias-tengler: For value types we would always create the cursor from the first element matching the node value, when the same node value is present multiple times in the page...

PR #9393, opened 2026-03-18
Declare RabbitMQ exchange before publishing messages

ittennull:
RabbitMQ requires an exchange to exist in order to publish a message to it, otherwise it returns an error and the whole channel becomes unusable. The code uses ConcurrentDictionary to remember which exchanges were declared. It is faster than calling the function before each publish.

The problem that this PR fixes appears when an application starts publishing updates to graphql subscription before a client connects to it. Since the queue and the exchange were created only on consumer side, the producer always failed if it was too fast because it couldn't publish to a non-existent exchange. Even worse - the whole channel got closed and all consumers and producers, even those that had create-first-publish-second approach, started to fail.


michaelstaib (changes requested):
This actually introduces more issues to the implementation and also introduces channel handling that is explicitly called our in the RabbitMQ docs to not do. 

We will do a general review of the subscription providers after 16 is out.

ittennull:
Could you elaborate: what problems does it introduce and what exactly is advised not to do in the docs?

ittennull:
> This actually introduces more issues

If I know what you mean, I can probably fix it

PR #9506, opened 2026-04-07
[Fusion] Add defer support

michaelstaib:


src/HotChocolate/Fusion/src/Fusion.Execution.Types/Completion/CompositeSchemaBuilder.cs
```diff
@@ -130,6 +130,38 @@ private static CompositeSchemaBuilderContext CreateTypes(
             }
         }
 
+        // Register the @defer directive so the gateway's validation accepts it.
+        // The gateway manages @defer itself (it does not pass it to subgraphs).
+        if (!directiveDefinitions.ContainsKey(DirectiveNames.Defer.Name))
```
tobias-tengler: I think this should be a composition setting you can enable / disable, since your infrastructure might not support it and client developers should only be able to use this, if it has been intentionally enabled.

PR #9541, opened 2026-04-13
[Fusion] Do not apply "InferKeysFromLookups" to invalid lookup fields

glen-84:

- [Fusion] Do not apply "InferKeysFromLookups" to invalid lookup fields.


src/HotChocolate/Fusion/test/Fusion.AspNetCore.Tests/__snapshots__/SourceSchemaErrorTests.No_Data_And_Error_With_Path_For_Lookup_Field_NonNull_OnError_Null.yaml
tobias-tengler, line 1: @glen-84 I think you can remove this test. The nullable lookup case should already be covered. This tests something that would be rejected, before its fed into the gateway.

PR #9545, opened 2026-04-13
Suppress ReSharper nullability errors in Mocha

glen-84:

- Suppress ReSharper nullability errors in Mocha.


glen-84:
`@Copilot What do you think about your own review here?`

I think the review was directionally thoughtful, but not well calibrated for this PR.

The core idea in the comments on [ReceiveEndpoint.Source](https://github.com/ChilliCream/graphql-platform/pull/9545#discussion_r3074816319) and [DispatchEndpoint.Destination](https://github.com/ChilliCream/graphql-platform/pull/9545#discussion_r3074816346) was reasonable: if something is genuinely nullable in supported use, the type system should say so. But in this case it pushed too hard toward “make the API nullable” and didn’t weigh the cost of widening a public contract just to silence a ReSharper warning. For a PR titled “Suppress ReSharper nullability errors in Mocha,” that’s too big a design leap.

What I’d say now is:

The review found a real tension between lifecycle state and nullability annotations.
It overstated the best fix. Targeted suppression was a better answer than changing public nullability.
I’d downgrade it from a design concern to a tooling/lifecycle note: “the analyzer can’t model this lifecycle cleanly; localized suppression is acceptable here.”
So my take is: useful instinct, wrong recommendation.

PR #9598, opened 2026-04-25
Use Options pattern for openapi and mcp

PascalSenn:
TODO:
- Add abstraction projects for openapi/mcp
- Expose `AddNitroOpenApiDefinitionStorage` as public `AddNitro`
- Remove `OpenApiTransportSetup` in favor of direct instantiations in manager


tobias-tengler:
For this to work, we would have to mirror a `AddOpenApi` on the GraphQL builders like with MCP, right?
Otherwise the necessary services do not get registered. Currently this is done by the Nitro packages calling `AddOpenApiDefinitionStorage` which registers the necessary services internally.
If we add that, we probably also don't need `OpenApiTransportSetup`.

We should also update the Map* and other methods to produce user-friendly messages like `Call AddMcp() ...` etc.
The current code already has some for the `AddOpenApiAspNetCoreServices` but this is something internal, that doesn't help the user...

PR #9758, opened 2026-05-21
[Fusion] Add LookupMustHaveArgumentsRule

tobias-tengler:


src/HotChocolate/Fusion/test/Fusion.Composition.Tests/SchemaComposerTests.cs
```diff
@@ -5,6 +5,45 @@ namespace HotChocolate.Fusion;
 
 public sealed class SchemaComposerTests
 {
+    // Regression: a @lookup field with no arguments used to crash composition with an
+    // IndexOutOfRangeException deep inside the selection-set rewriter. It must now fail with a
+    // proper LOOKUP_MUST_HAVE_ARGUMENTS composition error instead.
```
glen-84: I'm not a fan of these regression comments (which Claude often adds). It may be relevant now, but in future it doesn't really matter what "prompted" this fix, just that we have coverage for it.

PR #9765, opened 2026-05-22
Fix ExpressionHasher constant collision

apereiratl:
VisitConstant was not overridden, so predicates that differ only in their constant values hashed identically, collapsing aliased [UseFiltering] DataLoader branches into one.


src/GreenDonut/test/GreenDonut.Data.Tests/Internal/ExpressionHasherTests.cs
```diff
@@ -150,14 +186,14 @@ public static void BufferResize_Add_Expression()
         // arrange
         var hasher = new ExpressionHasher();
         var initialBufferSize = hasher.InitialBufferSize;
-        var expression = Expression.Lambda<Func<Entity1>>(Expression.Constant(new Entity1())); // translated to 8 bytes
+        var expression = Expression.Lambda<Func<int>>(Expression.Constant(0)); // translated to 31 bytes
```
tobias-tengler: Why did you have to change this?

tobias-tengler (commented):
Thanks for the contribution! :)

Where did you encounter this issue?
Could you add a test in `src/HotChocolate/Data/test/Data.Tests/IntegrationTests.cs` that shows that the `ExpressionHasher` fix solves the higher level issue you were having?

src/GreenDonut/test/GreenDonut.Data.Tests/Internal/ExpressionHasherTests.cs
apereiratl: Expression.Lambda<Func<Entity1>>(Expression.Constant(new Entity1())) hashed to 8 bytes per Add whilst VisitConstant was a no-op, only "Lambda|;" reached the buffer. After the fix the same
  expression emits roughly 110 bytes, varying with the test class's full name. That broke the 8-byte loop and tied the byte count to namespace. Swapped in Expression.Constant(0): primitive,
  deterministic 31-byte output, independent of where the test lives. The test still asserts a single buffer doubling.

apereiratl:
Hi @tobias-tengler I've fixed all the review comments. Please re-review and let me know if there's anything else. Thank you.

PR #9771, opened 2026-05-22
[Docs] Update v15 to v16 migration docs for service provider

DanielRose:
Fixes registration code example using `GetRequiredService`, to instantiate the correct class.

Makes it clearer that `AddApplicationService` always needs to be added, and the `AddDiagnosticEventListener` examples are two variants for the same thing.

Adds an example for an indirect/hidden registration (in this case logging).


tobias-tengler (approved):
Thanks! Could you make the same change in the Fusion migrate-from-15-to-16.md guide?

DanielRose:
@tobias-tengler I changed the Fusion docs as well.

PR #9786, opened 2026-05-27
[Fusion] Extend federation-gateway-audit satisfiability tests

glen-84:
## Summary
- Extend the federation-gateway-audit-derived section of `SatisfiabilityValidatorTests` with 16 suites across the `@override`, `@provides`, `@requires`, entity-call, and schema-shape clusters: 13 pass, 3 are committed as `[Fact(Skip = …)]`.
- Lookups follow one rule: an explicit `@lookup` is added for every type with a resolvable `@key` in the audit subgraph, since the composite-schemas spec has no implicit reference resolution (Federation's implicit `_entities` resolver becomes an explicit `@lookup`).
- SDLs are adapted to the composite-schemas spec: `@requires(fields:)` becomes argument-level `@require(field:)` (including `path<ConcreteType>.field` type conditions), and `@override(from:)` uses the helper's uppercase schema names.
- The 3 skipped suites are valid supergraphs in the audit project that hit current engine gaps (circular `@requires`, parent entity calls, and object-to-union widening at merge). Each quotes its blocker and asserts the intended `IsSuccess`, so it re-activates once the gap closes.

## Test plan
- `dotnet test src/HotChocolate/Fusion/test/Fusion.Composition.Tests --filter "FullyQualifiedName~SatisfiabilityValidatorTests"` — 61 passed, 3 skipped, 0 failed on net8.0, net9.0, and net10.0.

### DIFF
--- a/src/HotChocolate/Fusion/src/Fusion.Composition/Logging/LogEntryCodes.cs
+++ b/src/HotChocolate/Fusion/src/Fusion.Composition/Logging/LogEntryCodes.cs
@@ -70,5 +70,5 @@ public static class LogEntryCodes
     public const string RootSubscriptionUsed = "ROOT_SUBSCRIPTION_USED";
     public const string SpecifiedByUrlMismatch = "SPECIFIED_BY_URL_MISMATCH";
     public const string TypeKindMismatch = "TYPE_KIND_MISMATCH";
-    public const string Unsatisfiable = "UNSATISFIABLE";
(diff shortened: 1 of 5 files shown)


glen-84:
Closing: opened from the wrong branch (gai/rename-unsatisfiable-query-path-code, whose rename work already merged via #9780). Reopening from gai/audit-derived-satisfiability-tests.

PR #9810, opened 2026-06-01
Fix double-encoding of escaped string variables in multipart uploads

BickelLukas:
## Summary

When a GraphQL multipart request (`graphql-multipart-request-spec`) is processed, `HttpMultipartMiddleware` rewrites the `operations` JSON to inject file references in place of the `null` placeholders. During this rewrite, any string variable that was *not* a file was copied using `writer.WriteStringValue(reader.ValueSpan)`.

`ValueSpan` returns the **raw, still-escaped** JSON bytes, while `WriteStringValue(ReadOnlySpan<byte>)` treats its input as unescaped text and escapes it again. The result is that every escape sequence in a string variable was doubled.

For example, a variable with the value `test " test` (JSON: `"test \" test"`) was rewritten to `"test \\\" test"`, so the resolver received `test \" test` instead of `test " test`. The same applied to any backslash, newline, unicode escape, etc.

## Fix

Decode the value before writing it back, but keep a zero-allocation fast path when no escaping is involved:

```csharp
case JsonTokenType.String:
    if (reader.ValueIsEscaped)
    {
        writer.WriteStringValue(reader.GetString());
    }
    else
    {
        writer.WriteStringValue(reader.ValueSpan);
    }
    break;
```

- Escaped values are decoded with GetString() so the writer re-escapes them correctly.
- Unescaped values still take the raw-span path, avoiding the string allocation in the common case.

## Tests
Added `Upload_File_With_Escaped_String_Variable_Is_Not_Double_Encoded`, which sends a file upload alongside a string variable containing an escaped quote and asserts the resolver receives the original, correctly-decoded value. Verified on net8.0, net9.0, and net10.0.

### DIFF
--- a/src/HotChocolate/AspNetCore/src/AspNetCore.Pipeline/HttpMultipartMiddleware.cs
+++ b/src/HotChocolate/AspNetCore/src/AspNetCore.Pipeline/HttpMultipartMiddleware.cs
@@ -340,7 +340,14 @@ public sealed class HttpMultipartMiddleware : HttpPostMiddlewareBase
                 break;
 
             case JsonTokenType.String:
-                writer.WriteStringValue(reader.ValueSpan);
+                if (reader.ValueIsEscaped)
+                {
+                    writer.WriteStringValue(reader.GetString());
+                }
+                else
+                {
+                    writer.WriteStringValue(reader.ValueSpan);
+                }
                 break;
 
             case JsonTokenType.Number:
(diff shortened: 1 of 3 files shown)


michaelstaib:
Thanks for this @BickelLukas! Since this PR is from an org-owned fork (maintainer edits aren't available), I've carried your fix forward in #9839, which adds the Fusion `FileEntryBuilder` / `InMemorySourceSchemaClient` paths and a `JsonWriter`-level fix (framing escaped values on write so they pass through zero-copy). Continuing review there, with credit to your original work.

BickelLukas:
Thanks @michaelstaib looking forward to this fix getting merged and released!

PR #9813, opened 2026-06-01
Refresh NuGet package icons and consolidate Directory.Build.props

rstaib:
### Summary

This PR refreshes the NuGet package icons across all ChilliCream products and cleans up the `Directory.Build.props` hierarchy so that shared packaging metadata lives in one place, with each product only overriding what is specific to it.

### Icons

- Replaced the legacy `*-signet.png` icons with new, smaller `*-icon.png` variants:
  - `hotchocolate-icon.png`, `fusion-icon.png`, `greendonut-icon.png`, `nitro-icon.png`, `strawberryshake-icon.png`
  - Added icons for products that previously had none: `cookiecrumble-icon.png`, `mocha-icon.png`
- Removed the old signet assets (`greendonut-signet.png`, `hotchocolate-signet.png`, `nitro-signet.png`, `strawberryshake-signet.png`).
- Every product project now references its `*-icon.png` via `PackageIcon`.

### Build / packaging consolidation

- `src/Directory.Build.props` now holds all shared packaging metadata (authors, company, copyright, license, project/repository URLs, release notes, symbols).
- The icon `<None>` include is now driven by `$(PackageIcon)` with a single conditional item group, instead of hard-coding each signet path:
  ```xml
  <ItemGroup Condition="'$(PackageIcon)' != ''">
    <None Include="$(MSBuildThisFileDirectory)..\website\static\resources\$(PackageIcon)" Pack="true" PackagePath="$(PackageIcon)" Visible="false" />
    <None Include="$(MSBuildThisFileDirectory)..\LICEN*" Pack="true" PackagePath="" Visible="false" />
  </ItemGroup>
  ```
- Each product's `Directory.Build.props` now declares only its own `Product`, `PackageIcon`, and `PackageTags`:
  - Hot Chocolate, Fusion, Green Donut, Nitro, Strawberry Shake
  - New props files for **Cookie Crumble** and **Mocha**
- `HotChocolate.Templates.csproj` was aligned with the shared metadata (same authors/copyright/license/repo properties, new icon, LICENSE packed in).

### Metadata updates

- Copyright unified to **"Copyright © 2018 - present ChilliCream Inc."** across props files, templates, and the root `LICENSE`.

### Redirects (backward compatibility)

- nginx configs (`website/config/conf.d/default.conf` and `website-next/nginx/conf.d/redirects.conf`) now `permanent`-redirect old `*-signet.png` / `*-signet.svg` resource URLs to the new `*-icon.png` URLs.
- Fixed a pre-existing typo: `react-rsata` → `react-rasta`.

### Notes for reviewers

- Icon and LICENSE paths are resolved with `$(MSBuildThisFileDirectory)` so they work regardless of where a project sits in the tree (this was the path-resolution concern raised by Copilot).
- Packages without an icon (e.g. internal/test packages) are handled by the `'$(PackageIcon)' != ''` condition rather than failing the pack.


glen-84 (changes requested):
1. Why move it out of `src/Directory.Build.props`, which is already imported by most/all projects, and is the closest common root? We have a _lot_ of files in the repository root already. :slightly_smiling_face:
2. You seem to have removed the TargetFrameworks and other configuration (f.e. Roslynator) from `src/Directory.Build.props`.

I would suggest reverting to the `src/Directory.Build.props` configuration, and making the icon changes there.

rstaib:
> 1. Why move it out of `src/Directory.Build.props`, which is already imported by most/all projects, and is the closest common root? We have a _lot_ of files in the repository root already. 🙂
> 2. You seem to have removed the TargetFrameworks and other configuration (f.e. Roslynator) from `src/Directory.Build.props`.
> 
> I would suggest reverting to the `src/Directory.Build.props` configuration, and making the icon changes there.

1. Because of templates.
2. Yeah, saw that. Should be fixed now.

Directory.Build.props
```diff
@@ -0,0 +1,27 @@
+<Project>
+  <PropertyGroup>
+    <Authors>ChilliCream authors and contributors</Authors>
+    <Company>ChilliCream Inc.</Company>
+    <Copyright>Copyright &#169; 2026 ChilliCream Inc.</Copyright>
```
glen-84:
```suggestion
    <Copyright>Copyright &#169; 2018 ChilliCream Inc.</Copyright>
```

This should be the year of first publication. We could also use a range (`2018–2026`), but then we have to remember to keep it up-to-date, which is clearly not happening. 😅

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2024 ChilliCream Inc. (Michael & Rafael Staib)
+Copyright (c) 2026 ChilliCream Inc.
```
glen-84:
```suggestion
Copyright (c) 2018 ChilliCream Inc.
```

glen-84 (changes requested):

LICENSE
rstaib: ?

rstaib: The company did not exist in 2018

Directory.Build.props
rstaib: As far as I know the year is not even relevant. I just updated it so it does not look outdated.

LICENSE
glen-84: > Use the first-publication year (2018), not the company inception year. The two refer to different things and don't need to match.
> A copyright notice encodes two independent facts:
> 
> The year = when the work was first published. It describes the work, not the owner.
> The name = who owns the copyright now. It describes current ownership, not who owned it back then.
> 
> Under the (historical, now-optional) US notice rules in 17 U.S.C. § 401(b), the date element is specifically "the year of first publication of the work" — tied to the work, not to whoever happens to hold the rights today. That's exactly why you routinely see notices like "© 1998 BigCorp" on software BigCorp acquired in 2010: the year stays anchored to first publication, the name updates to the current owner. So "Copyright © 2018 ChilliCream Inc." is internally consistent even if the Inc. didn't exist in 2018 — it reads as "first published 2018; copyright held by ChilliCream Inc.," two true statements that don't have to share a date.

rstaib: changed it to `2018 - present` (this is how Microsoft does it in the VSCode repo)

Directory.Build.props
glen-84: Ya, technically it could be dropped (which was my first instinct), but I checked with Claude:

> Short version: today, neither form is legally required, and the difference between them has no effect on whether you hold copyright. But the year still does useful work, so the choice isn't purely cosmetic.
> 
> ...
> 
> What the year still does for you
> Even though it's optional, including the year and a proper notice has practical value:
> 
> - It marks the first publication date, which is one input into calculating the copyright term and can matter in disputes.
> - In the US specifically, a proper notice (which includes the year) can defeat an "innocent infringement" defense — an infringer can't credibly claim they didn't know the work was protected, which limits their ability to reduce damages. Omitting the notice weakens that.

Your (and Michael's) call. :slightly_smiling_face:

templates/server/HotChocolate.Template.Server.csproj
```diff
@@ -1,10 +1,12 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)..\..\'))" />
```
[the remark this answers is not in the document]
glen-84: @rstaib Ah, right, these are just templates, so they shouldn't have the import or PackageIcon.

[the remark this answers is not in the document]
rstaib: @glen-84 why?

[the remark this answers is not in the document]
rstaib: Templates are NuGet packages as well.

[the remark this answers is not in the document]
glen-84: No, they're not. There is only one package (`HotChocolate.Templates`).

`HotChocolate.Template.Server.csproj` etc. are templates (they even have placeholders like `0.0.0-placeholder`).

These project files are output when running `dotnet new ...`.

Side note: Since there's only one project left, we could reconsider whether the root-level `Directory.Build.props` is really necessary, or if we could just copy the props for this one case.

src/Mocha/src/Mocha.Abstractions/Mocha.Abstractions.csproj
```diff
@@ -1,4 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)..\'))" />
```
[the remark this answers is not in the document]
glen-84: How were these resolved? These imports are likely redundant, since the parent is imported automatically. I think that they should be removed.

glen-84 (approved):

PR #9836, opened 2026-06-03
**Improve handling of duplicate types across schema files to avoid HC…

andreashellquist:
# Summary of Changes

I've identified and fixed the `HC0065` bug in StrawberryShake's code generation pipeline. The issue manifested in two main areas.

https://github.com/ChilliCream/graphql-platform/issues/9837

---

## 1. `SchemaHelper.Load` — HC0065 Duplicate Type Name Fix

**File:**  
`src/StrawberryShake/CodeGeneration/src/CodeGeneration/Utilities/SchemaHelper.cs`

### Root Cause

When multiple schema files contain overlapping type definitions (common with Hygraph/CMS schemas where glob patterns may pick up multiple files, or when schemas are split across files), `SchemaHelper.Load` previously called:

```csharp
builder.AddDocument(document);
```

for **each** non-extension file independently.

This caused the Hot Chocolate type system to register the same type name multiple times, triggering the `HC0065` (*duplicate type name*) error.

Additionally, documents that contained **both** extension nodes and regular type definitions were treated entirely as extension files, causing the regular type definitions in those mixed files to be silently ignored.

#### Merge and Deduplicate Type Definitions

All non-extension definitions from all schema files are now merged into a single deduplicated document before calling `builder.AddDocument()`.

- Deduplication is keyed by type name.
- Uses `TryAdd(...)`.
- First occurrence wins.

#### Correctly Process Mixed Files

Files containing both:

- extension nodes
- regular type definitions

are now processed correctly, ensuring both categories are included.

#### Deduplicate Scalar Registrations

Added a `HashSet<string>` to track already-registered scalars and prevent duplicate `AddType(...)` calls.

#### Exclude Scalars from the Merged Document

Scalar type definitions are already registered through:

```csharp
builder.AddType(new AnyType(...))
```

Therefore, scalar definitions are excluded from the merged document to avoid potential registration conflicts.

---

## 2. `TypeDescriptorMapper.ExtractRuntimeType` — SS0009 Name Collision Resilience

**File:**  
`src/StrawberryShake/CodeGeneration/src/CodeGeneration/Mappers/TypeDescriptorMapper.cs`

### Root Cause

When generated C# output type names collided with existing registered type names (possible in large schemas with many similarly named types), the code would immediately throw a `TypeNameCollision` (`SS0009`) exception.

### Fix

Instead of failing immediately, the code now attempts alternative names by appending numeric suffixes:

- `_1`
- `_2`
- `_3`
- ...

This mirrors the behavior of:

```csharp
DocumentAnalyzerContext.ResolveTypeName(...)
```

and makes code generation more resilient when working with large or complex schemas.

---

## 3. Tests

**File:**  
`src/StrawberryShake/CodeGeneration/test/CodeGeneration.Tests/Utilities/SchemaHelperTests.cs`

### Added Test Cases

| Test | Description |
|--------|-------------|
| `Load_DuplicateTypesAcrossFiles_DoesNotThrowHC0065` | Verifies that overlapping type definitions across multiple schema files no longer trigger `HC0065`. |
| `Load_MixedExtensionAndDefinitionFile_ProcessesBoth` | Verifies that files containing both extensions and regular type definitions correctly process both. |

---

### Duplicate Types Across Schema Files

**Before**

- Duplicate type definitions caused `HC0065`.

**After**

- Duplicate definitions are deduplicated.
- First occurrence wins.

### Mixed Extension + Definition Files

**Before**

- Only extensions were processed.
- Regular type definitions were ignored.

**After**

- Both extensions and definitions are processed.

### Runtime Type Name Collisions

**Before**

- Name collisions immediately caused `SS0009`.

**After**

- Alternative names are automatically generated using numeric suffixes before failing.
- Code generation is more resilient for large schemas.


michaelstaib (changes requested):
This doesn't really solve the problem and could lead to a plethora of issues.

andreashellquist:
> This doesn't really solve the problem and could lead to a plethora of issues.

Thanks for the review. I'm primarily a user of the package who hit this bug directly when trying to use it — I don't have deep knowledge of the codebase architecture, just gave copilot a chance to sort it out. Happy to close this PR if the team prefers to solve it differently. The issue and repro repo are there for reference. I've reverted the TypeDescriptorMapper change. Could you clarify what you see as the proper approach? Should the deduplication happen upstream in the CLI download step, or would you prefer a validation-with-clear-error rather than silent dedup? Happy to adjust

Issue #9817, opened 2026-06-02
include directive with argument from variable not working with selection through data loader in v16


IharYakimush:
### Product

Hot Chocolate

### Version

16.0.11

### Link to minimal reproduction

https://github.com/IharYakimush/include-directive-bug/tree/main/ConsoleApp1

### Steps to reproduce

run reproduction project to execute following query and check console output
```
query ($value: Boolean!){ 
    t1:test{
        id
        name
    }
    t2:test @include(if: true){
        id
        name
    }
    t3:test @include(if: $value){
        id
        name
    }
}
```

### What is expected?

```
{
  "t1": {
    "id": "1",
    "name": "Test Entity"
  },
  "t2": {
    "id": "1",
    "name": "Test Entity"
  },
  "t3": {
    "id": "1",
    "name": "Test Entity"
  }
}
```

### What is actually happening?

```
{
  "t1": {
    "id": "1",
    "name": "Test Entity"
  },
  "t2": {
    "id": "1",
    "name": "Test Entity"
  },
  "t3": null
}
```

### Relevant log output

```shell

```

### Additional context

works as expected in v15

IharYakimush:
related issue: https://github.com/ChilliCream/graphql-platform/issues/9880

PR #9882, opened 2026-06-10
Fix QueryContext selector dropping fields under variable @include/@skip

### LINKED ISSUE #9880: QueryContext selector drops fields gated by a variable @include/@skip directive (incomplete fix for #9066)
### Product

Hot Chocolate

### Version

16.1.3

https://github.com/lemol/hotchocolate-querycontext-variable-include-repro

(self-contained, no database required; `dotnet run`)

### Steps to reproduce

[#9066](https://github.com/ChilliCream/graphql-platform/issues/9066) / [#9135](https://github.com/ChilliCream/graphql-platform/pull/9135) made the `QueryContext<T>` selector respect `@include` / `@skip`, but only for **literal** directive values. When the directive condition is a **variable**, the field is dropped from the selector regardless of the variable's runtime value.

Clone the reproduction above and run `dotnet run`. It builds an in-process schema with a `QueryContext<Item>` resolver that applies the selector to an in-memory `IQueryable` via `.With(context)`, then executes the same `name` selection five ways. The single source row always has `Name = "real-value-from-source"`.

### What is expected?

Every selection asks for the `name` field to be **kept** (`@include(if: true)` and `@skip(if: false)`), so `name` should be `"real-value-from-source"` in all five cases — matching the behavior of `[UseProjection]` + `IQueryable<>`.

When the directive uses a **variable**, the field is dropped from the `QueryContext<>.Selector` and comes back as the CLR default (`null`), even though the field is present in the response shape:

```
A. unconditional                   "name": "real-value-from-source"   ✅
B. literal  @include(if: true)     "name": "real-value-from-source"   ✅
C. variable @include(if: $v=true)  "name": null                       ❌
D. literal  @skip(if: false)       "name": "real-value-from-source"   ✅
E. variable @skip(if: $v=false)    "name": null                       ❌
```

Only the **literal** directive cases (B, D) work — the fix in #9135 handles literal `BooleanValueNode` conditions but not `VariableNode` conditions. Since GraphQL clients (e.g. Apollo) almost always emit `@include(if: $someVar)` rather than a literal, real-world queries consistently hit this and silently receive default values for any projected field placed under a conditional directive.

### Relevant log output

```shell

```

### Additional context

Reproduced on both `16.1.3` (latest stable) and `16.1.4-p.3` (preview). Appears to be an incomplete fix for #9066 — the variable-valued directive case is not covered.

lemol:
Fixes #9880.

## Problem

The `QueryContext<T>` selector drops a field from the projection when that field is gated by a **variable** `@include(if: $var)` / `@skip(if: $var)` directive — regardless of the variable's runtime value. The field is present in the response shape but comes back as the CLR default (`null`, `0`, empty array, …).

`@include`/`@skip` with **literal** boolean values work correctly (fixed in #9135); only the variable form is affected. Since clients such as Apollo almost always emit `@include(if: $someVar)` rather than a literal, real-world queries consistently hit this.

```graphql
query ($v: Boolean!) {       # $v = true
  items { id name @include(if: $v) }
}
# name is null, even though the source row has a value
```

Minimal reproduction: https://github.com/lemol/hotchocolate-querycontext-variable-include-repro

## Root cause

`HotChocolateExecutionSelectionExtensions.AsSelector` decided whether to bypass the cached selector and apply the runtime include flags based on `selection.IsConditional` — i.e. whether the **projected field itself** carries a directive. In practice the directive sits on a **child** field, so the projected field is never conditional, the cached `includeFlags = 0` expression is reused, and `CollectSelections` excludes the conditional child via `IsIncluded(0)`.

The `..._Respects_Variable_Include_Directive_Across_Requests` test added in #9135 used empty data for the `withName = true` case, so the dropped field never surfaced.

## Fix

Track whether any conditional selection participates while building the selector (`ConditionalTracker`, threaded through `SelectionExpressionBuilder` and marked in `CollectSelections`), and cache the built expression **only when none did**. Cached selectors therefore stay independent of the runtime include flags and are always safe to reuse; conditional projections are rebuilt per request from the current flags.

This is correct at any nesting depth (the tracker is marked wherever a conditional selection is collected in the subtree) and preserves caching for non-conditional operations.

## Tests

Adds `QueryContext_Selector_Projects_Field_When_Variable_Include_Is_True`, which selects a field via `@include(if: $withName)` with `$withName = true` over real data and asserts the field is projected and returned. It fails on `main` and passes with this change. The existing `QueryContext_Selector_*` tests and the full `HotChocolate.Data.Tests` suite continue to pass.

### DIFF
--- a/src/HotChocolate/Core/src/Execution.Projections/Extensions/HotChocolateExecutionSelectionExtensions.cs
+++ b/src/HotChocolate/Core/src/Execution.Projections/Extensions/HotChocolateExecutionSelectionExtensions.cs
@@ -78,16 +78,22 @@ public static class HotChocolateExecutionSelectionExtensions
         this Selection selection,
         ulong includeFlags)
     {
-        var isConditional = selection.IsConditional;
-
         // we first check if we already have an expression for this selection,
-        // this would be the cheapest way to get the expression.
-        if (!isConditional && TryGetExpression<TValue>(selection, out var expression))
+        // this would be the cheapest way to get the expression. A cached expression is
+        // only ever stored when the projected selection set (at any depth) contained no
+        // conditional (@include/@skip) selection, so it is independent of the runtime
+        // include flags and always safe to reuse here.
+        if (TryGetExpression<TValue>(selection, out var expression))
         {
             return expression;
         }
 
-        // if we do not have an expression we need to create one.
+        // we do not know up front whether the projection depends on the runtime include
+        // flags (a conditional selection can sit arbitrarily deep in the selection set),
+        // so we build the expression while tracking whether any conditional selection
+        // participated and only cache the result when none did.
+        var tracker = new SelectionExpressionBuilder.ConditionalTracker();
+
         // we first check what kind of field selection we have,
         // connection, collection or single field.
         var flags = selection.Field.Flags;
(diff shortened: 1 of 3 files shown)


tobias-tengler:
Hey :) Thanks for working on this. I already started working on a fix for this last week, I just need some inputs from another team member and then your issue should be resolved: https://github.com/ChilliCream/graphql-platform/pull/9818
You can subscribe to the existing issue to get notified once it's fixed: https://github.com/ChilliCream/graphql-platform/issues/9817

lemol:
This is awesome! Thank you @tobias-tengler !

PR #9987, opened 2026-06-25
Add scheduling support to the Mocha in-memory transport

glen-84:
## Summary

- Mocha's in-memory transport now honors scheduled delivery: a message dispatched with a future `ScheduledTime` (via `ScheduleSendAsync`/`SchedulePublishAsync`, a saga's `Timeout()`, or delayed redelivery) is delivered at its scheduled time instead of immediately.
- Adds an in-memory `IScheduledMessageStore` and a background `IHostedService` worker, wired into `AddInMemory` through Mocha's generic scheduling middleware (`UseSchedulerCore`). Cancellation is supported: a pending timeout is cancelled when a saga reaches a final state. Scheduled messages are held in process (non-durable); for durability use `UsePostgresScheduling()`.
- Updates the saga integration test harness to start the hosted scheduler worker so `Timeout()` sagas are driven by real scheduling.
- Updates the Mocha v16 docs (sagas, scheduling, in-memory transport) to reflect that in-memory scheduling is on by default, supports cancellation, and is non-durable.

## Test plan

- Store unit tests and end-to-end tests (via `FakeTimeProvider`): a scheduled message is delivered after its due time and not before, cancellation prevents delivery, and a scheduled message's body and headers are preserved until delivery.
- Full Mocha in-memory transport suite green (206 tests).
- Mocha saga suite green across net8.0–net11.0.


glen-84:
Closing in favor of #10119, which re-implements in-memory transport scheduling against the reworked scheduling subsystem introduced by #10036 (`ScheduledMessageStoreRegistration` + `ScheduledMessageStoreResolver`).

That rework landed after this PR was opened and made this branch's approach obsolete: this PR wired a plain `IScheduledMessageStore` singleton via `UseSchedulerCore()` and relied on `SupportsSchedulingNatively`, both of which no longer exist. #10119 registers the in-memory store per transport through the new resolver model, gives the in-memory worker a dedicated scheduler signal, and carries the docs and test updates forward.

Issue #9908, opened 2026-06-12
Source Generator does not set MethodInfo or ParameterInfo


leniency:
### Product

Hot Chocolate

### Is your feature request related to a problem?

Methods registered through the source generator do not have the `MethodInfo` or any `ParameterInfo` set on the `ObjectFieldConfiguration` or any of their `ArgumentConfiguration`. When then inspecting configurations in a TypeInterceptor, fields and arguments are missing information.

This information used to be available under HotChocolate 15, and it is available for fields and arguments _not_ being registered through the source generator, leading to a discrepancy in handling between those two paths.

### Sample Repo

I have a sample repository up here demonstrating the issue.
https://github.com/leniency/HotChocolate16Collections

- [CustomDirectiveInterceptor](https://github.com/leniency/HotChocolate16Collections/blob/master/HotChocolate16Collections/CustomDirectiveInterceptor.cs) this is a sample interceptor that applies a testing attribute at various locations. The FIELD_DEFINITION and ARGUMENT_DEFINITION are where the issue arises - `Member` and `Parameter` properties are null on the corresponding configuration objects.

### Use Cases

Adding custom directives based on attributes. I use a TypeInterceptor to check for various attributes to apply custom directives. Not all of these can be modified to inherit from `DescriptorAttribute`.

- DataAnnotations - The various validation and display attributes (`StringLength`, `Range`, etc...) can be applied as directives for client-side validation and kept in sync with server logic.

### The solution you'd like

Add `MemberInfo` and `ParameterInfo` to the configuration objects in the source generator, allowing TypeInterceptors and any other consumers to have access to the original type information.

Or, just any other method to access all attributes applied to objects/fields/arguments/etc on the graph in something like the TypeInterceptor so that custom middleware and directives can be applied consistently.

leniency:
I took a look at the source generator a bit and am wondering if this is an acceptable fix:

https://github.com/leniency/graphql-platform/commit/9e1c91ac5ffa53b072465ecba8e4c35be9f35168

- I moved the `configuration.Member` section higher, removed the conditionals around it limiting it to only methods returning nodes or collection segments.
- From that, get the method parameters and set that as it generates the argument configurations.

michaelstaib:
I will have a look at it

leniency:
[...]

PR #10039, opened 2026-07-02
Fix custom parameter binding in source-generated resolvers

glen-84:
## Summary
- Custom parameter expression builders registered via `AddParameterExpressionBuilder` are now honored on the source-generated resolver binding path, matching the reflection path. `CustomParameterExpressionBuilder` participates in the source-gen `ParameterBindingResolver` (it now implements `IParameterBindingFactory`), like the built-in builders.
- Previously such a parameter was misclassified as a GraphQL input argument, so a source-generated (`partial`) `[QueryType]`/`[MutationType]`/`[ObjectType]` resolver taking a context-supplied parameter could fail schema build when that CLR type was not a valid input type.
- When a builder is registered with a `ParameterInfo` predicate (which the source generator cannot evaluate), schema build now fails with a clear, actionable error instead of silently misclassifying the parameter.
- Related to #9908 (both stem from the source generator not surfacing parameter metadata).

## Test plan
- Unit tests for `ParameterBindingResolver`: a custom builder is classified as `Custom` (not `Argument`), its binding executes the registered delegate, and a `ParameterInfo`-predicate builder throws for a matching parameter.
- End-to-end source-generator tests: a `partial [MutationType]` with a custom parameter builds and executes with the parameter injected and absent from the schema as an argument; an interface/base parameter matched by a predicate builder fails loudly.

### DIFF
--- a/src/HotChocolate/Core/src/Types/Internal/CustomParameterExpressionBuilder.cs
+++ b/src/HotChocolate/Core/src/Types/Internal/CustomParameterExpressionBuilder.cs
@@ -1,5 +1,7 @@
+using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using HotChocolate.Resolvers;
 
 namespace HotChocolate.Internal;
@@ -222,6 +224,11 @@ public class CustomParameterExpressionBuilder<TArg> : CustomParameterExpressionB
         public bool IsPure => isPure;
 
         public T Execute<T>(IResolverContext context)
-            => (T)(object)compiled(context)!;
+        {
+            // the binding is only created for a parameter of type TArg, so T == TArg here
+            Debug.Assert(typeof(T) == typeof(TArg));
+            var value = compiled(context);
+            return Unsafe.As<TArg, T>(ref value);
+        }
     }
 }
--- a/src/HotChocolate/Core/src/Types/Internal/CustomParameterExpressionBuilder.cs
+++ b/src/HotChocolate/Core/src/Types/Internal/CustomParameterExpressionBuilder.cs
@@ -8,9 +8,11 @@ namespace HotChocolate.Internal;
 /// A custom parameter expression builder allows to implement custom resolver parameter
 /// injection logic.
 /// </summary>
-public abstract class CustomParameterExpressionBuilder : IParameterExpressionBuilder
+public abstract class CustomParameterExpressionBuilder
+    : IParameterExpressionBuilder
+    , IParameterBindingFactory
 {
-    private readonly bool _isPure;
+    private protected readonly bool _isPure;
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


glen-84:
## Design options considered

**Root cause:** the source-generated resolver binding path (`ParameterBindingResolver`) narrows custom parameter builders to `IParameterBindingFactory`. `CustomParameterExpressionBuilder` (created by `AddParameterExpressionBuilder`) implements only `IParameterExpressionBuilder`, so it was dropped — both `GetBindingInfo` (kind) and `GetBinding` (runtime value) fell back to the default argument binding. The reflection path keeps the full builder list, which is why non-`partial` resolvers worked and HC15 (reflection-only) was unaffected.

**Option A — make `CustomParameterExpressionBuilder` a dual-interface builder (chosen).** Implement `IParameterBindingFactory` on it, mirroring every built-in builder that already implements both interfaces. Minimal and framework-native; two internal Data builders (`PagingArguments`, EF `ContextFactory`) had already hand-rolled this exact workaround, which confirmed the shape.

**Option B — widen `ParameterBindingResolver` to adapt any `IParameterExpressionBuilder` (rejected).** More literally "align with reflection", but infeasible: the base interface is `ParameterInfo`/`Expression`-based, and the source-gen runtime has neither a `ParameterInfo` (its `ParameterDescriptor` carries only name/type/nullability/attributes) nor an expression-compile step. Only `CustomParameterExpressionBuilder` can be bridged (it holds a compilable `Expression<Func<IResolverContext, TArg>>`), so the fix belongs on that type.

**Option C — surface `ParameterInfo` in the source generator (deferred).** The strategic fix for the shared root cause (see #9908), via emitted synthetic metadata or runtime reflection. Larger, and out of scope here. If it lands, the predicate limitation below disappears.

**Sub-decision — custom `Func<ParameterInfo, bool>` predicates.** The source generator cannot evaluate an arbitrary `ParameterInfo` predicate. Rather than silently approximate it with a type check, the build now fails loud with an actionable error (register with the type-based overload). The relevance guard uses `parameter.Type.IsAssignableFrom(typeof(TArg))`, so a predicate matching an interface/base parameter is caught too. Default (type-based) `AddParameterExpressionBuilder<T>(...)` registrations are fully supported.

glen-84:
Superseded by #10041, which fixes the same root cause with a cleaner, less invasive change: a single file, with explicit interface implementations on the generic `CustomParameterExpressionBuilder<TArg>`, so it avoids touching `ParameterBindingResolver`, `PagingArguments`, and `ContextFactory`. Closing in favor of it.

The only behavior this PR added that #10041 does not: a custom `Func<ParameterInfo, bool>` predicate builder fails loud with a clear error under the source generator (instead of silently falling through to a GraphQL argument). Leaving that here as a possible follow-up if guarding that case is considered worthwhile.

PR #10119, opened 2026-07-17
Add scheduled delivery to the Mocha in-memory transport

glen-84:
## Summary

- The Mocha in-memory transport now supports scheduled delivery out of the box: a message dispatched with a `ScheduledTime` (via `ScheduleSendAsync`/`SchedulePublishAsync`, or a saga's `descriptor.Timeout(...)`) is delivered at ~that time instead of immediately, and can be cancelled via the returned token. `AddInMemory()` registers a transport-scoped store plus a hosted delivery worker on the post-#10036 `ScheduledMessageStoreRegistration`/resolver model. Delivery is in-process, non-durable (scheduled messages are lost on restart), and best-effort (no retry).
- The worker owns a **dedicated** `ISchedulerSignal` rather than the shared singleton, because `MessageBusSchedulerSignal` is a single-waiter primitive (two workers on one instance would corrupt each other's wait). Hardening that signal's notify/wait handshake against lost wake-ups (notify-before-wait and stale-target-between-waits) also benefits the Postgres dispatcher, which shares the type.
- `Mocha.EntityFrameworkCore.Postgres.Tests` scheduling tests are re-pointed onto a bare in-memory transport so they still exercise the EF `UsePostgresScheduling()` fallback: with in-memory now having native scheduling, the previous "in-memory transport + EF fallback" pairing is obsolete. Docs updated: in-memory scheduling is native, on by default, cancellable, and non-durable, with Postgres-specific troubleshooting scoped accordingly.

## Test plan

- New in-memory scheduling suites (store unit tests + end-to-end with `FakeTimeProvider`): not-delivered-before-due, delivered-after-due, cancellation prevents delivery, body/headers preserved across pooled-context reuse, same-due-time ordering.
- Scheduler-signal race tests (notify-before-wait, stale-target-between-waits) added red→green.
- Saga timeout integration test updated for real deferred scheduling.
- Verified locally, all green: InMemory 844, `Mocha.Tests` 3940, Sagas 568 pass/4 skip, `Mocha.EntityFrameworkCore.Postgres.Tests` scheduling 52 (Squadron/Docker).


src/Mocha/src/Mocha.Transport.InMemory/Scheduling/InMemoryTransportScheduledMessageStore.cs
```diff
@@ -0,0 +1,181 @@
...
+        var token = Add(envelope, scheduledTime);
+
+        return new ValueTask<string>(token);
+    }
+
+    /// <summary>
+    /// Deep-copies the envelope and stores it under a new token. The stored copy is independent of
+    /// the caller's envelope, so mutating the original body or headers afterward does not affect it.
+    /// </summary>
+    internal string Add(MessageEnvelope envelope, DateTimeOffset scheduledTime)
```
PascalSenn: should be public i guess (th eclass is internal)

src/Mocha/src/Mocha.Transport.InMemory/Scheduling/InMemoryTransportScheduledMessageStore.cs
```diff
@@ -0,0 +1,181 @@
...
+    {
+        var id = Guid.NewGuid();
+        var copy = new MessageEnvelope(envelope) { Body = envelope.Body.ToArray() };
```
PascalSenn: we probalby should have a static method on messageenvelope for this copy. i think this could be relevant in other cases aswell. the clone of the body is and important detail here.

src/Mocha/src/Mocha.Transport.InMemory/Scheduling/InMemoryTransportScheduledMessageStore.cs
```diff
@@ -0,0 +1,181 @@
...
+    public MessageEnvelope Envelope { get; } = envelope;
+
+    public DateTimeOffset ScheduledTime { get; } = scheduledTime;
+}
+
+/// <summary>
+/// Orders entries by scheduled time, then by ID, so distinct entries with the same due time coexist
+/// in the <see cref="SortedSet{T}"/> (a total order is required, otherwise adds are dropped).
+/// </summary>
+internal sealed class ScheduledEntryComparer : IComparer<ScheduledEntry>
```
PascalSenn, line 156: is it really worth the complexity to make this ordered? like wouldnt a simple linked list be enough. the in memory scheduler is a fallback anyway no?

src/Mocha/src/Mocha.Transport.InMemory/MessageBusBuilderExtensions.cs
```diff
@@ -19,6 +25,28 @@ public static IMessageBusHostBuilder AddInMemory(
 
         busBuilder.ConfigureMessageBus(b => b.AddTransport(transport));
 
+        busBuilder.Services.TryAddSingleton(
+            sp => new InMemoryTransportScheduledMessageStore(
+                sp.GetService<TimeProvider>() ?? TimeProvider.System));
```
PascalSenn: i think there is a sp.GetTimeProvider()

PascalSenn: (also can be a static delegate)

src/Mocha/src/Mocha/Scheduling/MessageBusSchedulerSignal.cs
```diff
@@ -10,18 +10,23 @@ internal sealed class MessageBusSchedulerSignal(TimeProvider timeProvider)
 
     private DateTimeOffset _target = DateTimeOffset.MaxValue;
     private CancellationTokenSource? _delayCts;
```
PascalSenn, line 12: this race condition fix seems to be in the wrong place. isnt this an order of monitor acquisition issue?

glen-84: There's only one lock here (`_lock`), taken by both `Notify` and `WaitUntilAsync`, so there's no acquisition-order hazard between them. The race is a lost wakeup: the wake is edge-triggered (`_delayCts.Cancel()`), but the sleep (`Task.Delay`) has to run outside the lock, since we can't hold a lock across an `await`. So a `Notify` that fires when no CTS is armed (before a wait starts, or in the gap between two waits) cancels nothing, and the wakeup is lost.

That can't be closed by lock ordering; it needs level-triggered state that survives the gap and is checked under the lock before sleeping. `_notified` is exactly that pending-wakeup flag, and `_isWaiting` guards the `scheduledTime >= _target` early-skip so a stale target left behind by a completed wait can't swallow a notify. It's the standard condition-variable pattern, just hand-rolled because the wait is async rather than `Monitor.Wait`/`Pulse`.

Where your instinct is right: this primitive is subtle. The cleaner version bases it on a level-triggered async primitive (`SemaphoreSlim(0,1)`, a `Channel`, or an async manual-reset event), which makes the notify-before-wait case impossible by construction and lets `_notified` go away; the `_target` comparison stays only as an early-skip optimization. That's a broader change though, since this signal is shared with the EF Postgres dispatcher. Happy to do it here or as a follow-up.

src/Mocha/src/Mocha.Transport.InMemory/Scheduling/InMemoryTransportScheduledMessageStore.cs
glen-84: The part flagged as complexity is really just the `(ScheduledTime, Id)` comparer, which any duplicate-tolerant ordered structure needs. A sorted linked list wouldn't be simpler: we'd still need the ordering, plus a `Dictionary` alongside it for O(1) cancel-by-token (cancellation looks up by token, not by position). `SortedSet` + `Dictionary` gives O(log n) add/cancel/peek with true removal.

If the goal is less code and we accept worse worst-case perf, the genuinely simpler option is a plain `List<ScheduledEntry>` with a linear min-scan and linear cancel, which is fine for the low pending-count an in-memory transport typically holds. A linked list specifically doesn't buy anything over either.

I lean toward keeping `SortedSet` (correct, scales, comparer is small), but happy to drop to the `List` version if you'd prefer simplicity. Which way do you want it?

PR #10131, opened 2026-07-21
[Analyzer] Add source method links to generated DataLoaders

tobias-tengler:


michaelstaib:
Lets talk this through tomorrow. This moves the issue into the docs, but I think we can solve this differently by moving more code into the generated code.

PR #10164, opened 2026-07-29
Fix Parent requirements for colliding resolver fields

### LINKED ISSUE #10163: [Parent(requires:)] is silently ignored under [UseProjection] when another resolver field is named after the required property
### Product

Hot Chocolate

### Version

16.5.1

https://github.com/kavthedev/hc-parent-requires-repro

### Steps to reproduce

Clone the reproduction https://github.com/kavthedev/hc-parent-requires-repro and run `dotnet run` two queries, no database, no setup.

1. `Brand { Id, Name }` is exposed through two root fields: `viaQueryContext`
   (`QueryContext<Brand>` + `.With(query)`) and `viaUseProjection` (`[UseProjection]`).
2. `BrandExtensions.GetDisplayName` declares `[Parent(requires: nameof(Brand.Name))]`.
3. The same type also declares `GetName([Parent] Brand brand) => "static-value"`  a resolver
   whose GraphQL field name (`name`) matches the required property. It declares no requirement
   of its own.
4. Query `{ viaUseProjection { displayName } }`.
5. Comment out `GetName` (one line, marked in the file) and run again.

### What is expected?

`Name` is projected either way, so `displayName` returns `ACME` from both root fields with or
without the unrelated `GetName` resolver present. Adding a field to a type should not change what
another field's declared `[Parent(requires:)]` projects.

With `GetName` present, `{ viaUseProjection { displayName } }` returns the CLR default (`""`) the
repro prints it as `<EMPTY: Name was not projected>`. `Name` never enters the projection.

- Comment out `GetName` and `displayName` returns `ACME`. That one line is the whole difference.
- It is not limited to the colliding field: every field on the type requiring `Name` loses it.
  `GetName` declares no requirement itself, so it is enough that the field exists.
- Scoped to the collided property: a `requires: Id` on the same type keeps working while `name`
  collides.
- The trigger is the GraphQL field name, not the method name `[GraphQLName("name")]` on an
  unrelated method reproduces it, and renaming `GetName` to `label` does not.
- `QueryContext<Brand>` + `.With(query)` honors the requirement in every variant.
- `descriptor.Field(x => x.Name).IsProjected(true)` restores the `[UseProjection]` case.

No exception, no warning, no SDL difference. On EF Core the dropped column arrives as `null` on a
non-nullable member, so it usually surfaces as an unrelated-looking DataLoader null-key exception
far from the cause.

### Relevant log output

```shell
{ viaQueryContext { displayName } }    -> { "data": { "viaQueryContext": [ { "displayName": "ACME" } ] } }
{ viaUseProjection { displayName } }   -> { "data": { "viaUseProjection": [ { "displayName": "<EMPTY: Name was not projected>" } ] } }
```

### Additional context

Also reproduces on `16.6.0-p.6` (newest preview at the time of writing), so it is not fixed on main.
Verified on net10.0 with an in-memory `IQueryable`, and on EF Core + PostgreSQL where the generated
SQL shows it directly: without the colliding field the query selects the required column, with it the
column is absent from the SELECT list.

tobias-tengler:
Fixes https://github.com/ChilliCream/graphql-platform/issues/10163

### DIFF
--- a/src/HotChocolate/Core/src/Types/Execution/Processing/OperationCompiler.cs
+++ b/src/HotChocolate/Core/src/Types/Execution/Processing/OperationCompiler.cs
@@ -115,6 +115,7 @@ public sealed partial class OperationCompiler
                 compilationContext,
                 _optimizers.SelectionSetOptimizers,
                 parentIsInternal: false,
+                parentIsProjectionRequirement: false,
                 ref lastId);
 
             compilationContext.Register(selectionSet, selectionSet.Id);
@@ -207,6 +208,7 @@ public sealed partial class OperationCompiler
                 compilationContext,
                 optimizers,
                 selection.IsInternal,
+                selection.IsProjectionRequirement,
                 ref lastId);
             compilationContext.Register(selectionSet, selectionSet.Id);
             elementsById = compilationContext.ElementsById;
@@ -291,6 +293,7 @@ public sealed partial class OperationCompiler
         CompilationContext compilationContext,
         ImmutableArray<ISelectionSetOptimizer> optimizers,
         bool parentIsInternal,
+        bool parentIsProjectionRequirement,
         ref int lastId)
     {
         var i = 0;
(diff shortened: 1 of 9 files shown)


michaelstaib (approved):
LGTM

PR #10291, opened 2026-08-24
Fix defer parentNodeId resolution and pooled path aliasing in operation plan traces

rstaib:
Fixes two @defer-related bugs in Fusion's operation plan output (`extensions.fusion.operationPlan`).

## IncrementalPlan.ParentNodeId resolves to the wrong root node

When a deferred fragment was anchored under a field resolved by the second (or later) root fetch node, the plan reported the first root node as `parentNodeId`. All root nodes share `Target = $`, so the deepest-ancestor selection in `ResolveDeferParentNodeId` always kept the first one. A candidate now only qualifies if the first field segment of the defer path below its target is a response name in its root `ResultSelectionSet`; the deepest-target preference among qualifying candidates is kept, with a fallback to the previous behavior when no candidate qualifies (avoids turning exotic shapes into `IncrementalPlanParentNotFound`).

## ExecutionNodeTrace.VariableSets aliases pooled path arrays

Trace variable sets captured `CompactPath`s wrapping `int[]`s rented from the store's path pool. After a node's results merged, the arrays were returned and re-rented by a later node, so serializing the trace at request completion printed the later node's requirement paths under the earlier node's `variableSets`. When telemetry is collected, `TrackVariableValueSets` now stores detached copies of `Path` and `AdditionalPaths` (new `CompactPath.Detach()` / `CompactPathSegment.Detach()`); the pooled originals still flow through the request/merge path unchanged, so the hot path is unaffected.

## Tests

- Folded the repro into `DeferTests` as `Defer_Anchored_Under_Second_Root_Node_Should_Report_Correct_ParentNode` (snapshot now pins `parentNodeId: 2`) and added a three-root case pinning `parentNodeId: 3`.
- New `OperationPlanTelemetryTests` asserts an earlier lookup's trace keys are its own target paths after a dependent lookup builds deeper paths; both new tests fail without their fix.
- Full `Fusion.Execution.Tests` and `Fusion.AspNetCore.Tests` suites pass; only the intended snapshot values changed.


michaelstaib (changes requested):
I am working on changes ... so do not merge

PR #10295, opened 2026-08-25
Issue 9908 - Source generator - set configuration Member and Parameter properties.

leniency:
This is a fix for Issue #9908 .

Primary changes:

- src/HotChocolate/Core/src/Types.Analyzers/FileBuilders/TypeFileBuilderBase.cs
Source generator now sets the configuration member and parameter properties.

- src/HotChocolate/Core/test/Types.Analyzers.Integration.Tests/Issue9908Tests.cs
Unit test validating the change.

The rest is snapshot updates for the source generated code.


michaelstaib (changes requested):
Yes ... we fixed it .... and this is not the solution `  Writer.WriteIndentedLine("var parameters = (configuration.Member as global::System.Reflection.MethodInfo)?.GetParameters();");`

The issue is that the LLM goes off in the wrong direction here. We do explicitly do not want reflection here. There are already fallback mechanismns in place. Let me check if i pushed already the fix or if it local.

Issue #9550, opened 2026-04-15
Tight coupling between `Optional<T>` and GraphQL abstractions


flespi:
### Product

Hot Chocolate

### Is your feature request related to a problem?

This is more of a suggestion for improvement than a feature request.

The `Optional<T>` type is an excellent way to express whether or not a value is intended to be set, even when that value might be `null`. In my opinion, it's a useful primitive that, unfortunately, is not available as a built-in type in .NET (because `Nullable<T>` is not the same thing).

This type is primarily used in mutation definitions. But what if your mutation contract is defined as a command in the application layer or implemented differently elsewhere? In that case, the corresponding layer must depend on `HotChocolate`.

I understand that the type is defined in the `HotChocolate.Abstractions` package. However, that's the same place where types such as `StateKind`, `ModuleOption`, `GraphQLException`, `ExecutionResult` are defined, to name just a few. I believe the `Optional<T>` type is a much more fundamental primitive, one that goes beyond the scope of GraphQL abstractions.

Even moving it to the `HotChocolate.Primitives` package still gives the impression that this simple yet powerful primitive remains unnecessarily coupled to a GraphQL framework. 

### The solution you'd like

It might be a bit presumptuous of me to expect it to move to a completely separate namespace. I'd like to hear your thoughts.

michaelstaib:
I do get your ask. But I dislike from our side to introduce unbound primitives. Like just moving it into a new package that has no philosophy behind it feels wrong.

For us `IOptional` represents a GraphQL primitive. Not only for mutations but in general for GraphQL inputs.

In your case I probably would map the input to your command which can have a completely decoupled optional, if that is something that is useful in your application layer.

michaelstaib:
I am closing this issue for now. If there is more demand for this we can reevaluate this again.

PR #10317, opened 2026-08-30
Define `ChilliCream.Primitives` package

flespi:
## Summary

This PR introduces a new `ChilliCream.Primitives` package and moves `Optional<T>` out of `HotChocolate.Abstractions`.

### Changes

* Added the new `ChilliCream.Primitives` package.
* Moved `Optional<T>` and `IOptional` from `HotChocolate.Abstractions` to `ChilliCream.Primitives`.
* Added the new project to the solution structure.
* Updated `HotChocolate.Abstractions` to reference `ChilliCream.Primitives`.

### Motivation

`Optional<T>` is a general-purpose primitive for representing the distinction between an unspecified value and an explicitly specified value, including `null`.

Keeping it in `HotChocolate.Abstractions` forces application-layer code that uses `Optional<T>`—for example, command or mutation contracts—to depend on Hot Chocolate abstractions.

Moving it into `ChilliCream.Primitives` allows `Optional<T>` to be reused independently of GraphQL and Hot Chocolate while keeping the existing Hot Chocolate API intact.

This addresses the decoupling concern described in #9550.

### Note

The existing namespace of `Optional<T>` and `IOptional` has been preserved to avoid introducing unnecessary breaking changes.

This approach preserves backward compatibility for the current version without requiring consumers to update their existing `using` directives or fully qualified type references.


michaelstaib (changes requested):
I do get your ask. But I dislike from our side to introduce unbound primitives. Like just moving it into a new package that has no philosophy behind it feels wrong.

For us `IOptional` represents a GraphQL primitive. Not only for mutations but in general for GraphQL inputs.

In your case I probably would map the input to your command which can have a completely decoupled optional, if that is something that is useful in your application layer.

